PR #1, opened 2022-03-03
make Filter have a second argument for index

Bluskript:


samber:
Oh yes, definitely!

Bluskript:
@samber aaa i forgot to mention the module version needs to be bumped to that `go get` can fetch the new version

samber:
Yep!

Let me add this argument to all functions before release ;)

samber:
Done v1.1.0

PR #6, opened 2022-03-04
update map test and Shuffle

retornam:
- use EqualValues for map tests

- update shuffle to default from rand which uses Fisher-Yates [1]

[1] https://en.wikipedia.org/wiki/Fisher–Yates_shuffle


samber:
Reverted my PR, in order to accept yours ;)

Thanks for the contribution !

retornam:
@samber 👍🏾

PR #9, opened 2022-03-05
Update go.yml

retornam:
use unstavle 1.18.0-beta1 until golang 1.18 is officially released


samber:
Nice tip!

let's merge ;)

retornam:
@samber Thanks.  For the `golangci-lint` workflow do you want me to include the default `.golangci.yml` or do you have your own set of lint rules. I wanted to check before fixing that as well

PR #13, opened 2022-03-05
parallel: remove mutex from Map and Times

fsouza:
The slice is pre-allocated, a mutex is not needed.

### DIFF
--- a/parallel/slice.go
+++ b/parallel/slice.go
@@ -7,17 +7,14 @@ import "sync"
 func Map[T any, R any](collection []T, iteratee func(T, int) R) []R {
 	result := make([]R, len(collection))
 
-	var mu sync.Mutex
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


parallel/slice.go
```diff
@@ -7,17 +7,14 @@ import "sync"
 func Map[T any, R any](collection []T, iteratee func(T, int) R) []R {
 	result := make([]R, len(collection))
 
-	var mu sync.Mutex
 	var wg sync.WaitGroup
 	wg.Add(len(collection))
 
 	for i, item := range collection {
-		go func (_item T, _i int) {
+		go func(_item T, _i int) {
```
fsouza, line 14: includes some fixes by gofmt, let me know if you'd prefer me to remove those

samber:
oh yes!

Since `result` is allocated earlier, we do not need locks anymore...

I will run benchmarks again.

PR #17, opened 2022-03-07
update Uniq parameters receive multiple slice

xyctruth:


samber:
Why did you suggest this change?

IMO it could introduce misunderstanding of function output: does it remove duplicated argument or duplicated element of each slice ?

I suppose the best way to do it is to chain Uniq and Flatten: `lo.Uniq(lo.Flatten(s1, s2, s3))`.

PR #18, opened 2022-03-07
Improve `Map[T any, R any]` speed

quantumsheep:
After doing a little bit of testing it's clear that `append` do some additional processing. I got around 25-30% speed improvements with these modifications.


samber:
Nice!

PR #21, opened 2022-03-08
Add FlatMap

yusuke-yajima-hacomono:
* Add FlatMap    
  ex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap


samber:
Good one, thanks for the contribution.

PR #23, opened 2022-03-08
Add ContainsBy

wu-xian:


samber:
Looks good. Thanks for the contribution ;)

PR #24, opened 2022-03-08
readme has extra arg for lop.Map() example

docwhat:
Looks like a simple typo -- there is an extra argument to the `lop.Match()` example.


samber:
Oops. Thanks 👌

PR #27, opened 2022-03-09
Add MapAsync

wanglihui:
Add MapAsync to Support Async Map; like Promise.all


samber:
I think it overlaps with `lop.Map`.

https://github.com/samber/lo/blob/master/parallel/slice.go

I'm not sure returning an error in `Map` is a good idea anyway. IMO, using an Optional/Maybe data structure is better.

WDYT ?

PR #28, opened 2022-03-09
Add Drop, DropRight, DropWhile and DropRightWhile

Azer0s:
*) https://lodash.com/docs/4.17.15#drop
*) https://lodash.com/docs/4.17.15#dropRight
*) https://lodash.com/docs/4.17.15#dropWhile
*) https://lodash.com/docs/4.17.15#dropRightWhile

### DIFF
--- a/drop.go
+++ b/drop.go
@@ -1,5 +1,6 @@
 package lo
 
+//Drop drops n elements from the beginning of a slice or array.
 func Drop[T any](collection []T, n int) []T {
 	if len(collection) <= n {
 		return make([]T, 0)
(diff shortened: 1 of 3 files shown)


Azer0s:
Should I also edit README.md?

drop.go
```diff
@@ -0,0 +1,61 @@
...
+func DropWhile[T any](collection []T, predicate func(T) bool) []T {
+	result := []T{}
+
+	i := 0
+	for ; i < len(collection); i++ {
+		if !predicate(collection[i]) {
+			break
+		}
+	}
+
```
samber, line 25: Since you know the size of result, you should allocate it here, and use `result[j - i] = xxx` instead of `append`.

(prevent a little overhead)

samber (changes requested):
Yes please, add a mention of these functions in README.

Can you add a comment on top of functions plz?

Our linter is broken for now, since it is not compatible with go1.18. But my eyes are linting pretty well actually. 😅

Good contrib anyway :)

drop.go
Azer0s: Done

PR #29, opened 2022-03-10
Add Try, TryCatch, TryWithErrorValue and TryCatchWithErrorValue

Azer0s:


Azer0s:
Same question as here: https://github.com/samber/lo/pull/28#issuecomment-1063522649

samber:
@mattn I just saw your 👎 emoji above.

I'm curious. Why don't you like it? 🤔

@Azer0s please add comments and README ;)

try.go
```diff
@@ -0,0 +1,46 @@
+package lo
+
+func Try(callback func() error) (err bool) {
```
samber, line 3: I'm not a big fan of the `Try` function.

IMO we should only provide `TryWithErrorValue`. Go developers are encouraged to not ignore errors.

If you really want so, you can still ignore the returned value with `_`.

Also, having `TryWithErrorValue` named `Try` will promote it as the by-default function.

WDYT ?

try.go
```diff
@@ -0,0 +1,46 @@
...
+	}()
+
+	if callback() != nil {
+		err = false
+	}
+
+	return
+}
+
+func TryWithErrorValue(callback func() error) (err bool, errorValue any) {
```
samber, line 19: I would rather use the following prototype, more respectful of go standards:

```go
func TryWithErrorValue(callback func() error) (errorValue any, ok bool) {
```

(reversed order of returned value)

mattn:
> @mattn I just saw your 👎 emoji above.
> 
> I'm curious. Why don't you like it? 🤔

11 years ago, I wrote similar Go project. https://github.com/mattn/go-try
But this does not solve any problems.

try.go
Azer0s: I added it because I have this in some of my projects as a helper function. Yes I know, I shouldn't ignore error values but sometimes it's nicer to write 

```go
if !Try(someFunction) {
  panic("errorred!")
}
```

than 

```go
if err := someFunction(); err != nil {
  panic("errorred!")
} 
```

try.go
Azer0s: 👍

try.go
Azer0s: Originally this was meant for functions that return a lot of values that I ignore and only need the error status from but I'm still playing around with that. So instead of 

```go
if err, _, _, _ := someFunction(); err != nil {
  panic("errorred!")
} 
```

I could write

```go
if !Try(someFunction) {
  panic("errorred!")
}
```

than

samber: Yes, I see.

Good point.

try.go
Azer0s: Alright, done

try.go
Azer0s: I have done some refactoring. There are now some overload methods for Try

So for the case above (so 3 return values and one error value) you could just write

```go
if !Try4(someFunction) {
  panic("errorred!")
}
```

WDYT? @samber

samber:
Merging. 

i will add documentation :)

PR #31, opened 2022-03-10
Add Range

jsrdxzw:


util.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	result := make([]int, -elementNum)
+	for i := 0; i < -elementNum; i++ {
+		result[i] = -i
+	}
+	return result
+}
+
+// RangeFrom creates an array of numbers from start with specified length.
+func RangeFrom(start, elementNum int) []int {
```
samber, line 20: `func RangeFrom[T constraints.Integer | constraints.Float](start T, elementNum int) []T {`

util.go
```diff
@@ -0,0 +1,58 @@
...
+	result := make([]int, -elementNum)
+	for i := 0; i < -elementNum; i++ {
+		result[i] = start - i
+	}
+	return result
+}
+
+// RangeOpen creates an array of numbers (positive and/or negative) progressing from start up to, but not including end.
+// step is default 1 when set to zero or can not reach end.
+func RangeOpen(start, end, step int) []int {
```
samber, line 37: `func RangeOpen[T constraints.Integer | constraints.Float](start T, end T, step T) []T {`

samber (changes requested):

util.go
Azer0s: Hmm...I don't really think `RangeOpen` is a good name for this. Maybe something like `RangeSteps` or `RangeWithSteps`? I feel like `RangeOpen` isn't really descriptive enough. 

WDYT?

jsrdxzw: May be RangeWithSteps is appropriate for this 😂

util.go
```diff
@@ -0,0 +1,58 @@
+package lo
+
+// Range creates an array of numbers (positive and/or negative) with given length.
+func Range(elementNum int) []int {
+	if elementNum >= 0 {
```
Azer0s, line 5: Range can be simplified into:

```go
func Range(start, elementNum int) []int {
	result := make([]int, If(elementNum < 0, -elementNum).Else(elementNum))
	it := If(start <= elementNum, 1).Else(-1)

	for j, i := 0, start; i != elementNum; j, i = j+1, i+it {
		result[j] = i + start
	}

	return result
}
```

util.go
samber: RangeFromWithSteps ? 😆 

I like RangeWithSteps too

util.go
jsrdxzw: Ogg. Yeah, I forgot there is a condition function that can be used..

PR #32, opened 2022-03-11
Add Debounce

jsrdxzw:


retry.go
```diff
@@ -1,5 +1,26 @@
...
+	"sync"
+	"time"
+)
+
+type debounce struct {
+	after time.Duration
+	mu    sync.Mutex
+	timer *time.Timer
+}
+
```
samber, line 13: Could you please add a `Cancel()` function ?

README.md
```diff
@@ -795,6 +796,22 @@ iter, err := lo.Attempt(0, func(i int) error {
...
+```go
+var counter uint64
+f := func() {
+    atomic.AddUint64(&counter, 1)
+}
+debounce := newDebounce(100 * time.Millisecond)
+// in each loop, func given is only invoked once time
+for i := 0; i < 10; i++ {
+    debounce(f, f, f, f, f)
+    time.Sleep(200 * time.Millisecond)
```
samber, line 810: In this example, the sleep time should be less than debounce time.

samber: In order to simplify the example, I recommend using the following function:

```
f := func() {
    println("Called once every 100ms!")
}
```

Counter is not essential. Let's keep samples as short as possible!

README.md
```diff
@@ -795,6 +796,22 @@ iter, err := lo.Attempt(0, func(i int) error {
...
+creates a debounced instance that delays invoking func given until after wait milliseconds have elapsed.
+```go
+var counter uint64
+f := func() {
+    atomic.AddUint64(&counter, 1)
+}
+debounce := newDebounce(100 * time.Millisecond)
+// in each loop, func given is only invoked once time
+for i := 0; i < 10; i++ {
+    debounce(f, f, f, f, f)
```
samber, line 809: You chose to set callbacks when calling `debounce()`.

Why did you choose to not set a static list of callbacks from `NewDebounce()` ?

README.md
```diff
@@ -795,6 +796,22 @@ iter, err := lo.Attempt(0, func(i int) error {
...
 ```
 
+### Debounce
+creates a debounced instance that delays invoking func given until after wait milliseconds have elapsed.
+```go
+var counter uint64
+f := func() {
+    atomic.AddUint64(&counter, 1)
+}
+debounce := newDebounce(100 * time.Millisecond)
```
samber, line 806: newDebounce -> lo.NewDebounce

samber (changes requested):
Thanks for your contribution @jsrdxzw !

I wrote some comments about Debounce API and README.

README.md
jsrdxzw: what about this:
`d.Add(f2).Add(f2).Add(f2).Add(f2).Add(f2)`

README.md
jsrdxzw: Yeah, hh

retry.go
jsrdxzw: Get it, I added `cancel` feature just now

retry.go
```diff
@@ -1,5 +1,40 @@
...
+func (d *Debounce) Add(f func()) *Debounce {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.done {
+		return d
+	}
+	if d.timer != nil {
+		d.timer.Stop()
+	}
+	d.timer = time.AfterFunc(d.after, f)
```
Azer0s, line 24: Also, maybe a nitpick, but what about the invocation time of the function itself? So say the function takes 50ms to run and the debounce is set to 100ms. Will the function be invoked after 150ms again? Should this be run in a goroutine so we can make sure it can be run again after 100ms?

Azer0s (changes requested):

retry.go
Azer0s: What if I have something like `.Add(f).Add(f)` and f takes 120ms to run? I couldn't invoke f again but I should be able to.

jsrdxzw: > 
I think this debounce function will be executed once in set time(100ms), if the function is called again in set time(100ms), it will cancel the previous and reset. So no matter how much time the f is taken, it will be triggered **once** after specified time(100ms)
May be it is different from throttle..

samber:
What is the use case of adding a callback after Debounce initialization?

The Lodash specifications suggest a unique callback being "debounced" many times.

```go
cb := func() {
     // todo
}

d := lo.NewDebounce(100 * time.Millisecond, CB)

for i := 0; i < 3; i++ {
    d.Debounce()
    time.Sleep(50 * time.Millisecond)
}

d.Cancel()
```

This is the case I met the more in programming. Maybe I miss something, please let me know.

Azer0s:
> The Lodash specifications suggest a unique callback being "debounced" many times.

I agree, this looks nicer.
But even nicer, imo, would be to return a wrapped function and a cancelation token. So something like this:

```go
cb := func() {
     // todo
}

debounced, c := lo.NewDebounce(100 * time.Millisecond, CB)

for i := 0; i < 3; i++ {
    debounced()
    time.Sleep(50 * time.Millisecond)
}

c.Cancel()
```

I also think that we should think about certain edge cases as I described above

samber:
Good idea!

But why not going further?

```go
debounced, cancel := lo.NewDebounce(100 * time.Millisecond, CB)
debounced()
cancel()
```

jsrdxzw:
> Good idea!
> 
> But why not going further?
> 
> ```go
> debounced, cancel := lo.NewDebounce(100 * time.Millisecond, CB)
> debounced()
> cancel()
> ```

I just update my code as Lodash style, it now returns `cancel` token and wrap given function

samber (approved):

PR #34, opened 2022-03-11
Add paraller of filter

jaronnie:


parallel/slice.go
```diff
@@ -1,6 +1,35 @@
...
+)
+
+// Filter iterates over elements of collection, returning an array of all elements predicate returns truthy for.
+// `predicate` is call in parallel.
+func Filter[V any](collection []V, predicate func(V, int) bool) []V {
+	result := []V{}
+
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+	wg.Add(len(collection))
```
Azer0s, line 14: This is not really a smart thing to do as lock and unlock the mutex over and over again. Say we were to have 10 goroutines, each one taking longer than the previous one. Using lop.Filter would actually be slower because we have mutexes to lock, unlock and goroutines to create. 

Smarter would be a pool of workers with channels sending jobs to them. So create `runtime.NumCPU()`s amount of goroutines, send tasks to them, collect all results and proceed. This gives worse best case performance but way better average performance.

Azer0s (changes requested):

parallel/slice.go
jaronnie: ok, thanks, so we need lop.Filter? I will try to revise it again.

samber: We should put predicate results into a map, key being the item index.

Then we create the final slice to output.

samber: About pooling, I'm not even sure it is necessary, because Go already limits the number of concurrent goroutines.

For applications having very high load, pools would need to be shared globally, between every helper calls. So is it the job of this library to handle those cases?

WDYT?

Azer0s: Yes, the go runtime handles pooling in theory but consider that for every goroutine a new stack is created. Each goroutine takes minimum 2kb. For 1000 elements this is 2mb. That's absolutely insane. + It will probably be slower than just using a for loop. A faster solution would be to implement a task pool.  I am currently working on a lo feature that implements a task pool. If you can wait a day, you can use that instead :)

jaronnie: > 

```go
func Filter[V any](collection []V, predicate func(V, int) bool) []V {
	result := []V{}

	var wg sync.WaitGroup
	var mu sync.Mutex
	wg.Add(len(collection))

	for i, item := range collection {
		go func(_item V, _i int) {
			key := predicate(_item, _i)
			mu.Lock()
			if key {
				result = append(result, _item)
			}

			mu.Unlock()

			wg.Done()
		}(item, i)
	}

	wg.Wait()

	return result
}
```
I don't understand why we need put predicate results into a map. Is it ok?

Azer0s:
Also as of right now, the resulting slice is unfortunate out of order

jaronnie:
> 

yeah I know it, but we really need to make it sequential?

jaronnie:
> Also as of right now, the resulting slice is unfortunately out of order as goroutine execution isn't sequential

```go
func Filter[V any](collection []V, predicate func(V, int) bool) []V {
	var result []V
	seen := map[int]bool{}

	var wg sync.WaitGroup
	var mu sync.Mutex
	wg.Add(len(collection))

	for i, item := range collection {
		go func(_item V, _i int) {
			key := predicate(_item, _i)
			mu.Lock()

			// make predicate results into a map, key being the item idx
			seen[_i] = key
			mu.Unlock()

			wg.Done()
		}(item, i)
	}

	wg.Wait()

	for i, item := range collection {
		if seen[i] {
			result = append(result, item)
		}
	}

	return result
}
```

Finally we do a for loop to append results, but is it really that good? I feel a little stupid, but hard to find a great way to do that

Azer0s:
> > 
> 
> yeah I know it, but we really need to make it sequential?

Yes, of course!

Azer0s:
I'll create a PR for my task pool today and will propose a rewrite of this function

jaronnie:
> 


ok just looking at your research about task tool

parallel/slice.go
samber:
```go
func Filter[V any](collection []V, predicate func(V, int) bool) []V {
        tmp := make([]bool, len(collection))

        // map
	var wg sync.WaitGroup
	wg.Add(len(collection))
	for i, item := range collection {
		go func(_item V, _i int) {
			tmp[_i] = predicate(_item, _i)
			wg.Done()
		}(item, i)
	}
	wg.Wait()

        // reduce
	result := []V{}
	for i, item := range collection {
	        if ok, _ := tmp[i]; ok {
                        result = append(result, item)
                }
        }

	return result
}
```

It looks like a map/reduce jobs.

jaronnie: yeah, I think we can refer to https://github.com/kevwan/mapreduce library

Azer0s:
> > 
> 
> ok just looking at your research about task tool

https://github.com/samber/lo/pull/37

There you go. It's a bit faster than the current Map implementation. I think you can pretty much copy paste that implementation, change line 32 from task_pool.go
So instead of 
```go
result[v.i] = fn(v.val, v.i)
``` 
you'd have 
```go
if fn(v.val) {
    result = append(result, v.val)
}
```

Alternatively you can also just use my task pool like so:
```go
result := make([]T, 0)
NewTaskPool[T](c, 4, func(t T, _ int) *int {
    if filter(t) {
        result = append(result. t)
    }
    return nil
})
```

Again not sure how performance is with that, just try it out

jaronnie:
> > > 
> > 
> > 
> > ok just looking at your research about task tool
> 
> #37
> 
> There you go. It's a bit faster than the current Filter implementation. You can just use that.
> 
> ```go
> result := make([]T, 0)
> NewTaskPool[T](c, runtime.NumCPU()/2, func(t T, _ int){
>     if filter(t) {
>         result = append(result, t)
>     }
> })
> ```

ok let me look at your task pool

PR #35, opened 2022-03-11
Add Reject implementation

nitin1259:
Reject is opposite of filter - https://lodash.com/docs/4.17.15#reject


slice.go
```diff
@@ -240,3 +240,17 @@ func ToMap[K comparable, V any](collection []V, iteratee func(V) K) map[K]V {
 
 	return result
 }
+
+
+// Reject - the opposite of Filter, this method returns the elements of collection that predicate does not return truthy for.
+func Reject[V any](collection []V, predicate func(V, int) bool) []V {
+	result := []V{}
```
Azer0s, line 247: Wouldn't it be better to implement this as not filter? 

So something like:

```go 
notPredicate := func(val V, i int) bool {
    return !predicate(val, i)
}
return Map(collection, notPredicate)
```

Azer0s (changes requested):

README.md
```diff
@@ -456,6 +457,17 @@ l := lo.DropRightWhile[string]([]string{"a", "aa", "aaa", "aa", "aa"}, func(val
 // []string{"a", "aa", "aaa"}
 ```
 
+### Reject
+
+the opposite of Filter, this method returns the elements of collection that predicate does not return truthy for.
```
samber, line 462: the -> The

slice.go
samber: Yes!

I won't cost a lot more.

PR #37, opened 2022-03-13
Add task pool

Azer0s:
This is more of a prototype right now but it is already a bit faster than the normal lop.Map implementation.

image

Running on M1 Macbook Pro 2021

### DIFF
--- a/parallel/task_pool.go
+++ b/parallel/task_pool.go
@@ -1,0 +1,61 @@
+package parallel
+
+import (
+	"math"
+	"sync"
+)
+
+type job[T any] struct {
+	val T
+	i   int
+}
+
+func NewTaskPool[T, R any](source []T, poolSize int, fn func(T, int) R) []R {
+	result := make([]R, len(source))
+	cycles := int(math.Ceil(float64(len(source)) / float64(poolSize)))
+	n := len(source)
+	c := 0
+
+	wg := &sync.WaitGroup{}
+
+	jobChans := make([]chan job[T], poolSize)
+	for i := 0; i < len(jobChans); i++ {
+		jobChans[i] = make(chan job[T], cycles)
+		wg.Add(1)
+		go func(i int, in <-chan job[T]) {
+			for {
+				v, ok := <-in
+				if !ok {
+					wg.Done()
+					return
+				}
+				result[v.i] = fn(v.val, v.i)
+			}
+		}(i, jobChans[i])
+	}
+
+	for i := 0; i < cycles; i++ {
+		n -= poolSize
+		batchSize := 0
+		if n < 0 {
+			batchSize = n + poolSize
+		} else {
+			batchSize = poolSize
+		}
+
+		for j := 0; j < batchSize; c, j = c+1, j+1 {
+			jobChans[j] <- job[T]{
+				val: source[c],
+				i:   c,
+			}
+		}
+	}
+
+	for i := 0; i < len(jobChans); i++ {
+		close(jobChans[i])
+	}
+
+	wg.Wait()
+
+	return result
+}
--- a/parallel/task_pool_test.go
+++ b/parallel/task_pool_test.go
@@ -1,0 +1,33 @@
+package parallel
+
+import (
+	"testing"
+)
+
+func BenchmarkNewTaskPool(b *testing.B) {
+	l := 100000
+	c := make([]int, l)
+	for i := 0; i < l; i++ {
+		c[i] = i
+	}
+
+	for n := 0; n < b.N; n++ {
+		NewTaskPool(c, 4, func(t, _ int) float64 {
+			return float64(t)
+		})
+	}
+}
+
+func BenchmarkMap(b *testing.B) {
+	l := 100000
+	c := make([]int, l)
+	for i := 0; i < l; i++ {
+		c[i] = i
+	}
+
+	for n := 0; n < b.N; n++ {
+		Map[int, float64](c, func(t, _ int) float64 {
+			return float64(t)
+		})
+	}
+}


Azer0s:
Currently I have a return slice and just map the values with a function, I will change this however to let the users manage return data themselves

Azer0s:
image

Alright, done. Made it a bit faster even

Azer0s:
I have done some benchmarks and will commit them later. As of right now, the task pool implementation is about 4x faster.

parallel/slice.go
```diff
@@ -1,98 +1,56 @@
...
-
-			result[_i] = res
-
-			wg.Done()
-		}(item, i)
-	}
-
-	wg.Wait()
-
+	NewTaskPool(collection, runtime.NumCPU()/2, func(v T, i int) {
```
samber, line 12: Why did you hard code the pool size instead of using a parameter to the function? 🤔

A default value could be useful as well:

```go
var DefaultPoolSize = runtime.NumCPU()/2
```

I'm also thinking about the case of i/o bound jobs. The number of parallel jobs won't depend on number of CPU ;)

parallel/task_pool.go
```diff
@@ -0,0 +1,58 @@
...
+	"math"
+	"sync"
+)
+
+type job[T any] struct {
+	val T
+	i   int
+}
+
+func NewTaskPool[T any](source []T, poolSize int, fn func(T, int)) {
```
samber, line 13: Good implementation

I think we should add the following code on top of your function:

```go
if poolSize > len(source) {
   poolSize = len(source)
}
```

It will limit pool size in cases where source collection is smaller than the number of workers.

samber (changes requested):
Thanks for your contrib @Azer0s !

I made a few comments

parallel/slice.go
```diff
@@ -1,98 +1,56 @@
...
+	NewTaskPool[T](collection, runtime.NumCPU()/2, func(v T, _ int) {
+		key := iteratee(v)
 
-			mu.Unlock()
-			wg.Done()
-		}(item)
-	}
-
-	wg.Wait()
+		resultMu.Lock()
```
samber, line 44: I wonder if we could remove this lock as well, pipelining results into a channel and a single consumer.

I don't know the cost of mutex versus channel. Can you execute a benchmark?

Or a simpler implementation would be to write `iteratee` result into a slice and group later?

parallel/slice.go
Azer0s: Done

parallel/task_pool.go
Azer0s: Done

parallel/slice.go
Azer0s: Tried both. This is the fastest way for now. Unless we get to know the groups upfront, this is the fastest way.

Azer0s:
Refactored everything according to comments. Shall I add some doc in README? @samber

samber:
Yes please!

samber (approved):

Azer0s:
@samber Done. Could you look through the documentation I added?

Azer0s:
@samber I added another helper function called With. This is useful for the situations described in the readme and for some situations with the taskpool.

Azer0s:
please merge @samber

Bin-Huang:
func Map[T any, R any](collection []T, iteratee func(T, int) R, options ...*ParallelOption) []R {
}

Azer0s:
@Bin-Huang Ok I found a solution that imo is a bit more elegant. 
Kind of a compromise between your and my solution.

Could you take a look? @Bin-Huang @samber

Bin-Huang:
@Azer0s @samber
I also wrote some code this weekend, maybe we can share more ideas on the code.

https://github.com/samber/lo/pull/81

parallel/slice.go
```diff
@@ -1,134 +1,88 @@
...
-func Map[T any, R any](collection []T, iteratee func(T, int) R) []R {
-	result := make([]R, len(collection))
-
-	var wg sync.WaitGroup
-	wg.Add(len(collection))
-
-	for i, item := range collection {
-		go func(_item T, _i int) {
-			res := iteratee(_item, _i)
+var DefaultPoolSize = runtime.NumCPU() / 2
```
Bin-Huang, line 9: Is it `runtime.NumCPU()` can be 1, then `DefaultPoolSize` will be zero?

Bin-Huang: I know some use cases will cause confusing behavior about default pool size. 

Let says someone wants to request 10 URLs in parallel on a device had 2 CPUs, he will write some code just like that:

```go
responses := parallel.Map(URLs, request)
```

In this case, requesting a URL is only an IO-intensive task, so for the best performance and efficiency, it should requesting 10 URLs at the same time, just like the code above looks like for the package user.  

But DefaultPoolSize is working internally and it is `1` (half of the CPU number), so actually, only one requesting task is started at the same time. And in this case, someone will confused about **why the code will be slower 10x times**.

The task pool can improve performance only if the developer knows what they doing, and adjusts the pool size thoughtfully for the special scene. A global default pool size can't improve performance in any scene, and often it's worse (if it can, the Golang GMP will make it internally).

I recommend removing the default pool size, and only using the task pool when the concurrency limit option is set.

yjc567:
I think this PR is very useful, are we planing to merge it? 😂

brettmorien:
Would love to see this feature. Bit of a blocker to have unlimited goroutines getting spun up when using parallel functions.

kcmvp:
These methods are very useful, when those are be merged ?

kcmvp:
@Azer0s thank your great contribution very much! I like these feature as well. Could please resolve the conflicts and merge it ? @samber

Bin-Huang (approved):

Azer0s:
@kcmvp @Bin-Huang is this still relevant? If so I can fix the conflicts and merge

fmonpelat:
Hi there, sorry to bring this back but im need some of the options to send to map, @Azer0s could you fix the conflicts so it could be merged?

Issue #16, opened 2022-03-06
Allow restricting concurrency for parallel op's


freakynit:
Current implementation spawn as many go-routines as passed collection size (for example as passed to `Map`) method in `parallel` package. While this is perfectly okay to do for up to a few thousand go-routines, it could become problematic for larger values. For example, for a `map`ping operation that makes any networking call, this is almost equivalent to `DOS`sing the end service.

Suggestion is to allow configuring concurrency for such parallel operations. Default would be equal to `runtime.NumCPU()`, whereas, if anything <= 0 is specified, it would automatically switch to current default behavior. Anything greater than 0 would override previous behaviors specified.

Sample implementation with above in mind for `Map` method can be found here (this diff is against current implementation): https://www.diffchecker.com/4pLAUeMh

Let me know if this looks good.  I can work on the PR.

Best wishes...

Azer0s:
I implemented this in https://github.com/samber/lo/pull/37

PR #43, opened 2022-03-16
Add MapWithError

NilsJPWerner:
This pr adds both lo.MapWithError and lop.MapWithError.

This is very useful if your internal function returns an error and you want to short circuit the Map. Currently you can keep track of an error outside of Map, however it won't short circuit the mapping and any subsequent errors will overwrite each other. This is a common situation when doing anything io related.

To implement the parallel version I included the `golang.org/x/sync/errgroup` package in order to use `errgroup.Group` which implements go routine cancellation.

This is what lo.MapWithError looks like:
```go
x, err := lo.MapWithError[int64, string]([]int64{1, 2, 3, 4}, func(x int64, _ int) (string, error) {
    if x == 5 {
        return nil, errors.New("x is 5")
    }
    return strconv.FormatInt(x, 10), nil
})
// []string{"1", "2", "3", "4"}
// nil
```

Here is a less contrived example of lop.MapWithError that shows parallel http requests that could error.
```go
x, err := lop.MapWithError[string, string]([]string{"a", "b", "c"}, func(x string, _ int) (string, error) {
    resp, err := http.Get("https://example.com/" + x)
    if err != nil {
        return "", err
    }

    body, err := ioutil.ReadAll(resp.Body)
    if err != nil {
        return "", err
    }

    return string(body), nil
})
```


NilsJPWerner:
Also might be worth incorporating the suggestion from https://github.com/samber/lo/issues/16 since there are no limits to the number of goroutines that can be started.

NilsJPWerner:
Hey @samber do you have time to look at this?

dolmen:
`parallel.MapWithError` should be in separate PR.

parallel/slice.go
```diff
@@ -1,9 +1,13 @@
...
-import "sync"
+import (
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+)
 
 // Map manipulates a slice and transforms it to a slice of another type.
-// `iteratee` is call in parallel. Result keep the same order.
+// `iteratee` is called in parallel. Result keep the same order.
```
dolmen, line 10: This change should be in a separate MR.

NilsJPWerner:
I removed the errGroup import and the comment fix

NilsJPWerner:
Bump again @samber

tooilxui:
I also agree with the value of this PR

YamiOdymel:
It will be the best Christmas gift if it gets merged.

tooilxui:
> It will be the best Christmas gift if it gets merged.

So true

deregtd:
:sadge:

I've been thinking about forking lo to get MapWithError for a while now.  Keep hoping this will get merged but looks pretty dead.

denis-bc:
@samber could you please review this. 🙏 We all see community needs it! 🙂

samber:
✅ Done in https://github.com/samber/lo/pull/823

Please upgrade to v1.53.0

PR #46, opened 2022-03-18
feat: ✨  add Except function

### LINKED ISSUE #44: Feat: Except function
This is a request for a new function: `Except`. This would work similar to `Difference` except that it documents (assumes) that list1 is a superset of list2. This is the logical inverse of the `Intersect()`.

I understand that Difference actually has this behavior:

https://go.dev/play/p/WxfAXrK70wX

From a user experience perspective, it does a bit more than I need (I have to throw away the second array). And then there's this:

image

I'm happy to create a PR. Please let me know if you would like this. I believe that `Except` would be slightly faster, and with a slightly smaller memory footprint as well.

Thank you for this awesome library!!

ghostsquad:
Except is the logical inverse of intersect

resolves #44


intersect.go
```diff
@@ -95,6 +95,40 @@ func Difference[T comparable](list1 []T, list2 []T) ([]T, []T) {
...
 
+// Except returns distinct values from list1 except if that value is also in list2.
+func Except[T comparable](list1 []T, list2 []T) []T {
+	// bool is 1byte
+	// we are tracking what we want to _keep_
+	// with an initial value of false
+	// since we want to return distinct values only
+	// once we've seen the value, we'll change it to true
+	// and skip over future occurrences
+	keepSeen := make(map[T]bool)
```
samber, line 106: Value will never be `true`

can be changed by: `keepSeen := make(map[T]struct{})`

intersect.go
```diff
@@ -95,6 +95,40 @@ func Difference[T comparable](list1 []T, list2 []T) ([]T, []T) {
 	return left, right
 }
 
+// Except returns distinct values from list1 except if that value is also in list2.
```
samber, line 98: Can you explain why you chose to return distinct values? 🤔

samber (changes requested):
I made a comment and added a commit for documentation ;)

intersect.go
ghostsquad: done

ghostsquad:
@samber this PR has been updated. Thanks for your patience!

ghostsquad:
I fell off the bandwagon here. I can resolve conflicts here if you, @samber, still believe this is a worthy contribution. Let me know.

ghostsquad:
`Without` does what `Except` does, so I'm closing this PR.

PR #47, opened 2022-03-18
fix: 🐛  make intersect return unique values

### LINKED ISSUE #45: Bug: Intersect should only return distinct values
Lodash describes this operation as returning unique entries
https://lodash.com/docs/#intersection

Also described in SQL:
https://www.postgresql.org/docs/9.4/queries-union.html

> INTERSECT returns all rows that are both in the result of query1 and in the result of query2. Duplicate rows are eliminated unless INTERSECT ALL is used.

> EXCEPT returns all rows that are in the result of query1 but not in the result of query2. (This is sometimes called the difference between two queries.) Again, duplicates are eliminated unless EXCEPT ALL is used.

https://docs.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-except-and-intersect-transact-sql?view=sql-server-ver15

ghostsquad:
Lodash describes this operation as returning unique entries
https://lodash.com/docs/#intersection

Also described in SQL:
https://www.postgresql.org/docs/9.4/queries-union.html

INTERSECT returns all rows that are both in the result of query1 and in the result of query2. Duplicate rows are eliminated unless INTERSECT ALL is used.

https://docs.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-except-and-intersect-transact-sql?view=sql-server-ver15

resolves #45


wirekang:
Why this PR is not merged?

ghostsquad:
I guess I need to update it. @samber is this a candidate for review if I would to resolve the conflicts?

ghostsquad:
@samber is this a candidate for review if I would to resolve the conflicts?

ghostsquad:
it looks like intersect might have been updated with a similar implementation. So this pull request is no longer needed.

PR #49, opened 2022-03-20
Use variadic parameters in Min and Max

hanlins:
Greetings and thanks to this amazing project! I used `Min` and `Max` quite frequently and I found it annoying that I have to put the values into an array even when I got two elements. From my previous exp with c++, variadic parameters is used in `max` signature so developer can enter any number of parameters without creating an array/slice on themselves (see [ref](https://www.cplusplus.com/reference/algorithm/max/)).

The usage for `Max` could be:
```golang
max := lo.Max(1, 2, 3)
```
If developer wants to find the max element from an array, they can do:
```golang
var array := []{1, 2, 3}
max := lo.Max(array...)
```


NilsJPWerner:
This is a great suggestion!

NilsJPWerner (approved):

NilsJPWerner:
It might be worth replacing the Zip2 through Zip9 functions with a single variadic version as well.

Issue #56, opened 2022-03-21
Add function for maps to receive slice of values filtered by given keys


metalrex100:
Such function could be named like `ValuesF` which means `Values Filtered` or may be you suggest another naming.

The implementation is just an extension of the std lib `slices.Values` function:

```
ValuesF[K comparable, V any](m ~map[K]V, keys []K) []V {
	r := make([]V, 0, len(m))
	for k, v := range m {
		if slices.Contains(keys, k) {
			r = append(r, v)
		}
	}
	return r
}
```

metalrex100:
I may suggest PR if needed.

samber:
Hi @metalrex100 and thanks for the contribution.

I would rather choose `ValuesFilter` for a more explicit naming

But I think this helper does too many things.

Why not implementing:
- OmitByKeys(map, keysToRemove)
- OmitByValues(map, valuesToRemove)
- OmitBy(map, predicate)
- PickByKeys(map, keysToKeep)
- PickByValues(map, valuesToKeep)
- PickBy(map, predicate)

Your `ValuesF` function could be replaced by:

```go
subset := lo.PickByKeys(map[string]int{"foo": 1, "bar": 2, "baz": 3}, []string{"foo", "baz"})
values := lo.Values(subset)
```

WDYT ?

metalrex100:
@samber I thing that `PickByKeys` is a more clear naming for such functionality, indeed. Will change it.

samber:
I justed added:

- OmitByKeys(map, keysToRemove)
- OmitByValues(map, valuesToRemove)
- OmitBy(map, predicate)
- PickByValues(map, valuesToKeep)
- PickBy(map, predicate)

PR #62, opened 2022-03-22
[Feature]: Added ValuesF helper

metalrex100:
Closes https://github.com/samber/lo/issues/56

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ go.work
 cover.out
 cover.html
 .vscode
+
+.idea/
--- a/README.md
+++ b/README.md
@@ -540,12 +540,12 @@ values := lo.Values[string, int](map[string]int{"foo": 1, "bar": 2})
 // []int{1, 2}
 ```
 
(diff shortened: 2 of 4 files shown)


metalrex100:
@samber I renamed helper to `PickByKeys` as you suggested in #56

samber:
Can you adapt the helper to return a map instead of a list of values please?

```go
func PickByKeys[K comparable, V any](in map[K]V, keys []K) []V {
	r := make(map[K]V)
	for k, v := range in {
		if Contains(keys, k) {
			r[k] = v
		}
	}
	return r
}
```

Then your initial usecase could be solved by `lo.Values(lo.PickByKeys( ... ))`

metalrex100:
@samber done.

Also a small tip about tests: `testify` package has a func called `ElementsMatch`. It checks that both slices have same elements and ignores their order. So you can use it in tests for comparing slices instead of `sort+Equal`.

samber (approved):

PR #76, opened 2022-03-24
Add ReductionSteps

olavfosse:
This function is very handy for quickly seeing what is happening
inside a Reduce invocation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Supported helpers for slices:
 - Filter
 - Map
 - FlatMap
-- Reduce
+- Reduce / ReductionSteps
 - ForEach
 - Times
 - Uniq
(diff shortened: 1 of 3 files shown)


samber:
Why not using a debugger in that case ?

samber:
Why not using a debugger in that case ?

olavfosse:
The same reason one would use `fmt.Print` when one can inspect a value in the debugger; perhaps you don't have a debugger available, perhaps you want to serialize the value, perhaps you want to log it or perhaps you want to do any of the things one might want to do with a slice, aside from just inspecting it in the debugger.

This is a more-or-less standard function available in many programming languages, including [Wolfram Language](https://reference.wolfram.com/language/ref/FoldList.html), [Clojure](https://clojuredocs.org/clojure.core/reductions), [Haskell](https://hackage.haskell.org/package/base-4.6.0.1/docs/Prelude.html#g:15), [F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html#scan) and [OCaml](https://ocaml.org/api/Seq.html#VALscan) which in my opinion is proof of its usefulness.

Issue #77, opened 2022-03-25
Add concurrency limit option to package `lo/parallel`


Bin-Huang:
Sometimes if the collection is too big, it will make too many goroutines in one time when using `parallel` methods. That will significantly increase resource usage, especially when the `iteratee` function is computationally-Intensive, io-Intensive, or both.

An example that requests every URL in a huge list, will start so many requests at the same time:
```
hugeUrlList := []string{ /*...*/ }
parallel.ForEach(hugeUrlList, request)
```

I think there should have an option to set the max concurrency count, which means up to configured number `iteratee` calling  at the same time:

```
hugeUrlList := []string{ /*...*/ }
parallel.ForEach(hugeUrlList, request, parallel.Option().Concurrency(100))
```

I will enjoy implementing it.

Azer0s:
This is already implemented in #37

PR #81, opened 2022-03-29
feat: add max concurrency limit option to pkg parallel

Bin-Huang:
Sometimes if the collection is too big, it will make too many goroutines in one time when using `parallel` methods. That will significantly increase resource usage, especially when the `iteratee` function is computationally-Intensive, io-Intensive, or both.

An example that requests every URL in a huge list, will start so many requests at the same time:
```go
hugeUrlList := []string{ /*...*/ }
parallel.ForEach(hugeUrlList, request)
```

There should have an option to set the max concurrency count, which means up to configured number `iteratee` calling  at the same time:

```go
hugeUrlList := []string{ /*...*/ }
parallel.ForEach(hugeUrlList, request, parallel.Option().Concurrency(100))
```

Set max concurrency limit with optional parameters.
```go
handler := func(x int64, _ int) string {
    return strconv.FormatInt(x, 10)
}
lop.Map([]int64{1, 2, 3, 4}, handler, lop.Option().Concurrency(20))
```

Related issue https://github.com/samber/lo/issues/77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 @samber: I sometimes forget to update this file. Ping me on [Twitter](https://twitter.com/samuelberthe) or open an issue in case of error. We need to keep a clear changelog for easier lib upgrade.
 
+## 1.27.1 (2022-08-15)
+
+Bugfix:
+
+- Removed comparable constraint for lo.FindKeyBy
+
 ## 1.27.0 (2022-07-29)
 
 Breaking:
--- a/find.go
+++ b/find.go
@@ -98,7 +98,7 @@ func FindKey[K comparable, V comparable](object map[K]V, value V) (K, bool) {
 }
 
 // FindKeyBy returns the key of the first element predicate returns truthy for.
-func FindKeyBy[K comparable, V comparable](object map[K]V, predicate func(K, V) bool) (K, bool) {
+func FindKeyBy[K comparable, V any](object map[K]V, predicate func(K, V) bool) (K, bool) {
 	for k, v := range object {
 		if predicate(k, v) {
 			return k, true


FlandiaYingman (approved):
lgtm

Bin-Huang:
> lgtm

Thank you for your encouragement and keep me motivated to do the rest, such as docs, more methods updating...

Bin-Huang:
### Benchmark

previous(master):
```
goos: darwin
goarch: amd64
pkg: github.com/samber/lo
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkMap/lo.Map-12         	      21	  53455758 ns/op	39998870 B/op	 1000001 allocs/op
BenchmarkMap/lop.Map-12        	       4	 276965265 ns/op	119999904 B/op	 3000006 allocs/op
BenchmarkMap/reflect-12        	       3	 424862161 ns/op	170326728 B/op	 4000042 allocs/op
BenchmarkMap/for-12            	      22	  51211756 ns/op	39998864 B/op	 1000001 allocs/op
PASS
coverage: 0.7% of statements
ok  	github.com/samber/lo	8.301s
```

current branch:
```
goos: darwin
goarch: amd64
pkg: github.com/samber/lo
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkMap/lo.Map-12         	      20	  54369975 ns/op	39998294 B/op	 1000001 allocs/op
BenchmarkMap/lop.Map-12        	       4	 283803768 ns/op	119998416 B/op	 3000005 allocs/op
BenchmarkMap/reflect-12        	       3	 428028874 ns/op	170326048 B/op	 4000042 allocs/op
BenchmarkMap/for-12            	      22	  50519951 ns/op	39998184 B/op	 1000001 allocs/op
PASS
coverage: 0.7% of statements
ok  	github.com/samber/lo	8.224s
```

Not much has changed

Bin-Huang:
If you have time please get me some reviews and suggestions? @samber 

Currently, I don't have any idea about performance improvement, and I think @Azer0s's work is great. Maybe the PRs have different purposes but just look like each other.

DillonStreator:
Thoughts on adjusting the API to use a more idiomatic option interface like `With...`.. i.e. `WithConcurrency`

I made some adjustments to this branch so lmk what you think @Bin-Huang 

diff of changes https://diffy.org/diff/1a10f430669df

Bin-Huang:
> Thoughts on adjusting the API to use a more idiomatic option interface like `With...`.. i.e. `WithConcurrency`
> 
> I made some adjustments to this branch so lmk what you think @Bin-Huang
> 
> diff of changes https://diffy.org/diff/1a10f430669df
> 
> **edit** also just created a branch on my fork for review https://github.com/DillonStreator/lo/tree/with-concurrency [DillonStreator@9519c4c](https://github.com/DillonStreator/lo/commit/9519c4c674046d9a61e4107b19fd4e2ac5e07f64)

@DillonStreator I really like it. I think it is better.

dolmen:
This implementation allows this:
```go
lop.Map(list, iteratee, lop.Option().Concurrency(20))
```

But the API would be simpler if it allowed this:
```
lop.Map(list, iteratee, lop.Concurrency(20))
```

This could be achieved using the [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) design pattern.

DillonStreator:
@dolmen the current implementation is still using the functional options pattern but it is (IMO) unnecessarily namespacing options via the `Options()` function.

Namespacing does seem important when dealing with a growing suite of top level APIs. I think the namespacing can be done in a simpler and more conventional manner by using the `With...` prefix on options. i.e. `lop.WithConcurrency(20)`. I proposed this change a few comments up in https://github.com/samber/lo/pull/81#issuecomment-1094085213.

Bin-Huang:
Now we can do something like this:
```go
lop.Map(list, transform, lop.WithConcurrency(20))
```
```go
option := lop.WithConcurrency(20)
```
lop.Map(list, transform, lop.WithConcurrency(20).WithOther("in furture"))

```

**Comment by Bin-Huang** [comment]
Any things above were implemented, already for review.

File: README.md [hunk 1]
```diff
@@ -208,6 +208,12 @@ lop.Map[int64, string]([]int64{1, 2, 3, 4}, func(x int64, _ int) string {
 // []string{"1", "2", "3", "4"}
 ```
 
+You also can control the conurrency limit by optional ParallelOption to limit the maximum number of concurrent `iteratee` goroutines running at the same time.
```
kallydev, line 211: Found few typos, conurrency => concurrency.

kallydev:
I merged #81 and #43 in my [fork](https://github.com/kallydev/lo/commit/9b7160680e30de1b1b669de7c791b3e3d8db99c2) and it worked great. I hope to see it in the main branch soon. ❤️ 

```go
if transactions, err = lop.MapWithError(transactions, func(transaction *model.Transaction, _ int) (*model.Transaction, error) {
	block, err := ethereumClient.BlockByNumber(ctx, big.NewInt(transaction.BlockNumber))
	if err != nil {
		return nil, err
	}

	transaction.Timestamp = time.Unix(int64(block.Time()), 0)

	return transaction, nil
}, lop.NewOption().WithConcurrency(MaxConcurrency)); err != nil {
	return nil, err
}
```

README.md
Bin-Huang: > Found few typos, conurrency => concurrency.

Fixed.

jcv-yang:
Good feature!
When will this be merged?

Bin-Huang:
> Good feature! When will this be merged?

@samber

PR #90, opened 2022-04-09
add parallel methods for maps

DillonStreator:
Adds parallel helper methods for maps:
- `ForEachKeyValue` call `iteratee` concurrently for each key value pair in `m` with no limit on concurrency
- `ForEachKeyValueMax` call `iteratee` concurrently for each key value pair in `m` with `max` limit on concurrency

### DIFF
--- a/parallel/map.go
+++ b/parallel/map.go
@@ -1,0 +1,40 @@
+package parallel
+
+import "sync"
+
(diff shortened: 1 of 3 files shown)


DillonStreator:
Assuming https://github.com/samber/lo/pull/81 gains traction, I can adjust these methods to support concurrency options

Issue #61, opened 2022-03-22
Suggestion: FindIndex function


jalavosus:
The `Find` function exists. The `IndexOf` function exists. What would be super helpful is a `FindIndex` function; similar in mechanics to `Find`, except it returns the slice index of the first slice member which satisfies the `Find` function (or -1 if none are found). 

Awesome project, by the way!

metalrex100:
@jalavosus I think the implementation may be done like this:
```
func FindIndex[T any](collection []T, predicate func(T) bool) (int, bool) {
	for index, item := range collection {
		if predicate(item) {
			return index, true
		}
	}
	return 0, false
}```

PR #91, opened 2022-04-11
Implement FindIndex

caalberts:
FindIndex searches an element in a slice based on a predicate and returns the index and true.
It returns -1 and false if the element is not found.

#61


samber:
thanks for the contrib.

FindLastIndex would be cool too!

I will add the README documentation for both ;)

Issue #99, opened 2022-04-20
Accept error and bool with Must()


mvrahden:
Additionally to error returns, you also often stumble upon boolean returns.

Fictional example:

```Go
func Parse(raw string) (out ConcreteType, ok bool)
```

It is possible to use type switches within generic function, so we could switch for error and bool alike.
Allowing the Must() function to become slightly more generic than it is right now without breaking anything.

wirekang:
I tried this
![image](https://user-images.githubusercontent.com/43294688/164486080-7e7ee7e1-14e1-446d-8da8-a7854f221a68.png)
And
![image](https://user-images.githubusercontent.com/43294688/164486196-35938731-ddc6-4ec8-b66e-7da4c22cf041.png)


 How about `MustOK(Parse(""))`?

samber:
`MustOk` is good!

mvrahden:
No need, just allow `any` instead of `error`. You throw away compile time type matching for runtime assertion. It's not a problem because people who will use Must() expect their code to `panic` at this point anyway.

wirekang:
> No need, just allow `any` instead of `error`. You throw away compile time type matching for runtime assertion. It's not a problem because people who will use Must() expect their code to `panic` at this point anyway.

I didn't think that far. Check out my PR. `Must*` panics when `err` is `error` or `false`. Is it ok to remove condition `err != nil`?

mvrahden:
Your PR looks good :) thank your for your support on this 💯 

P.S: Just out of curiosity, why is each test case wrapped into separate blocks? I'd probably place them into individual `t.Run`s instead and give more context via the description. It's probably just a question of preference, but I'm curios anyway

wirekang:
> P.S: Just out of curiosity, why is each test case wrapped into separate blocks? I'd probably place them into individual `t.Run`s instead and give more context via the description. It's probably just a question of preference, but I'm curios anyway

I copied and modified the existing `TestMustX` code. There is no other intention.

PR #100, opened 2022-04-21
feat: add MustOk

wirekang:
`error` can't become union type like `error |  bool` because `error` has method `Error() string`. So I added MustOk.  #99


samber:
@wirekang should we close this PR, since #103 supports both error and bool?

PR #104, opened 2022-04-22
fix: wrong examples of Clamp function in readme

bugstop:
the output in readme is wrong:
```go
func main() {
	r1 := lo.Clamp(0, -10, 10)
	// 10

	r2 := lo.Clamp(-42, -10, 10)
	// -42

	r3 := lo.Clamp(42, -10, 10)
	// 42

	fmt.Println(r1, r2, r3)
}
```
Actual: `0 -10 10`


samber:
gosh!

thanks ;)

PR #111, opened 2022-04-24
docs: 📝  move examples to godocs

ghostsquad:
added benefit of this change is that the examples also become tests!

This is not currently complete, but I wanted to push this PR to see what you think @samber.


ghostsquad:
this would also fix #112 as all examples become tests, and thus "fail" if they are wrong.

samber:
Hi @ghostsquad and thanks for the contrib!

That looks cool!

I made the first documentation quickly as I was not familiar enough with godoc. It gonna improve readability a lot :tada:.

Do you know if we can add a step in the CI pipeline for ensuring helpers have an example?

How would you work on this? A single big PR? Multiple iterations?

Can you move `SourceSample` and `IntnSample` to a separated PR please?

ghostsquad:
We could run the tests, export the result (seeing the test names) then run another test that asserts that there are examples for each function.

ghostsquad:
> Can you move `SourceSample` and `IntnSample` to a separated PR please?

Ya, not a problem. I did that because I realized that I would not be able to easily generate a deterministic example in the case that tests are run in parallel. That's the problem with package level global variables, including when setting `rand.Seed`.

ghostsquad:
hey, sorry, went radio silent for awhile. @samber, are you still interested in this feature? I can pick this back up if so.

PR #114, opened 2022-04-26
Add Async{0-1} to be able to handle function with no return

CorentinClabaut:


concurrency.go
```diff
@@ -8,3 +8,18 @@ func Async[T any](f func() T) chan T {
 	}()
 	return ch
 }
+
+// Async0 executes a function in a goroutine and returns a channel closed after execution.
+func Async0(f func()) chan struct{} {
+	ch := make(chan struct{})
+	go func() {
+		f()
```
samber, line 16: Why not sending a message into channel, like we did for `Async`?

```go
f()
ch <- struct{}{}
close(ch)
```

concurrency.go
```diff
@@ -8,3 +8,18 @@ func Async[T any](f func() T) chan T {
...
+		close(ch)
+	}()
+	return ch
+}
+
+// Async1 executes a function in a goroutine and returns the result in a channel.
```
samber, line 22: Add:

```go
// Async1 is an alias to Async.
```

samber (changes requested):

concurrency.go
CorentinClabaut: No particular reason.

That might be enough then WDYT?

```go
f()
ch <- struct{}{}
```

CorentinClabaut:
I'm actually thinking of adding functions Async{2-7} that would return tuples with the results of the function passed in parameter to AsyncX.

```go
func Async2[A any, B any](f func() (A, B)) chan Tuple[A, B] {
	ch := make(chan Tuple[A, B])
	go func() {
		ch <- T2(f())
	}()
	return ch
}
```

Then we could have:
```go
ch := lo.Async2(func() (int, string){  return 1, "2"})
```
Instead of:
```go
ch := lo.Async(func() lo.Tuple2[int, string] { return lo.T2(1, "2")})
```

It would make it simpler to just re-use existing functions directly as AsyncX parameter.

WDYT @samber

samber:
This contrib looks nice.

I prefer your last version, with `callback func() (A, B)` prototype.

Let's merge.

PR #117, opened 2022-04-27
feat: Add AwaitAll function

wjdqhry:


concurrency_test.go
```diff
@@ -26,3 +26,47 @@ func TestAsync(t *testing.T) {
 		is.Fail("Async should not block")
 	}
 }
+
+func TestAwait(t *testing.T) {
```
CorentinClabaut, line 30: I think this should be `TestAwaitAll` or the functions should be `AwaitX`

concurrency_test.go
```diff
@@ -26,3 +26,47 @@ func TestAsync(t *testing.T) {
...
+	is := assert.New(t)
+
+	aAsync := Async(func() int { return 1 })
```
CorentinClabaut, line 33: It might be best not to use `Async` for this test, that way, if Async is broken, this test would still pass.

You could just create channels in the main thread and send a value to it before to call AwaitAllX.

CorentinClabaut (commented):
That looks good. 

I was thinking of adding something like that too.

I have made some suggestions on the PR if you want. And I think you forgot to update the Readme.md

Do you think we should try to add a version with a timeout as well?

wjdqhry:
> That looks good.
> 
> I was thinking of adding something like that too.
> 
> I have made some suggestions on the PR if you want. And I think you forgot to update the Readme.md
> 
> Do you think we should try to add a version with a timeout as well?

I think set timeout is a really good idea. 

However, I wonder if the function should really handle it, given that it is the channel received as a parameter, not the function that causes the timeout. 

Also, I think of this function as syntax sugar for now.

I'll have to update the Readme.md file too!

samber:
Hi @wjdqhry and thanks for your first contribution.

Are these helpers even necessary? 🤔

wjdqhry:
Thank you for your comments!

Above all, I think it was effective in that I could play the role of Syntax Sugar through Generic.

PR #118, opened 2022-04-27
Implement Unpack{2->9}

### LINKED ISSUE #115: Add function UnpackT{2->9} to unpack tuples
It would be useful to have functions like that:

```go
func UnpackT2[A any, B any](tuple Tuple2[A, B]) (A, B){
    return tuple.A, tuple.B
}
```

That would allow to extract variables from tuples which would make the code more readable in some situations.

I can provide a PR for this.

CorentinClabaut:
closes #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ Supported math helpers:
 Supported helpers for tuples:
 
 - T2 -> T9
+- Unpack2 -> Unpack9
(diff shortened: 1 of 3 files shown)


samber:
Looks good ;)

`Unpack2` is better than `UnpackT2` 👌

Issue #123, opened 2022-05-05
Request: Chop(string) and Chomp(string)


mcandre:
Can we get some more common string operations, such as the classic chop (force one rune truncation) and chomp (remove trailing runes only on CR, LF, or CRLF match)?

wirekang:
I will make ruby version of those.

mcandre:
I already have a library for Go, called go-chop, that offers these.

But I wish to promote this functionality into a more well maintained Go library, such as lo.

PR #124, opened 2022-05-05
Add Chop #123

wirekang:
Add #123

### DIFF
--- a/string.go
+++ b/string.go
@@ -1,6 +1,9 @@
 package lo
 
-import "unicode/utf8"
+import (
+	"strings"
+	"unicode/utf8"
+)
 
 // Substring return part of a string.
 func Substring[T ~string](str T, offset int, length uint) T {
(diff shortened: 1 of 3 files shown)


samber:
Can you explain the use case please ?

string.go
```diff
@@ -28,3 +31,18 @@ func Substring[T ~string](str T, offset int, length uint) T {
 func RuneLength(str string) int {
 	return utf8.RuneCountInString(str)
 }
+
+// Chop returns str with the last character removed.
+// If str ends with \r\n, both characters are removed.
```
ccoVeille, line 36: Bit it could end with \n and it won't work?

What happens if the last character is space or tab, or multiple \n or \r\n ?

I'm unsure this behavior would be expected.

Also if str is "a\r\nb", what do you expect? "a", "a\r\n" ?

I mean right now you code trim \r\n or the last character.

I'm unsure I would use a method with such behavior. What is the need that you are trying to solve?

string_test.go
```diff
@@ -43,3 +43,14 @@ func TestRuneLength(t *testing.T) {
...
+
+func TestChop(t *testing.T) {
+	is := assert.New(t)
+
+	is.Equal("", Chop(""))
+	is.Equal("", Chop("a"))
+	is.Equal("a", Chop("ab"))
+	is.Equal("a", Chop("a\n"))
+	is.Equal("a", Chop("a\r\n"))
+	is.Equal("a\n", Chop("a\n\r"))
```
ccoVeille, line 55: Please add tests about the expected behaviors about the use cases I listed

ccoVeille:
I don't understand either.

But its 2 years old PR?

BTW, why are you apparently addressing a bunch of old things right now? Thanks

mattg-imprint:
The use case is unclear and the other comments haven't been addressed. It's been 2 years, at this point it seems fair to close this as "wontfix" or whatever label if you want.

PR #128, opened 2022-05-05
Add Bind #126

wirekang:
Add #126


samber:
Hi @wirekang and thanks for the contrib.

I will rename it to `Partial` and merge. ;)

PR #132, opened 2022-05-10
feat: Add ToAnySlice

mimol91:
Small helper function which returns slice with all elements mapped to any type.

It is common for other libraries to accept a variadic number of arguments of interface type. _ToAnySlice_ allows easily to convert slice with concrete type to any


wirekang:
How about `lo.ToAnySlice`?

samber:
Yes, `lo.ToAnySlice` looks better!

Let's merge ;)

Issue #136, opened 2022-05-18
Nth incorrect behavior - also, panic instead of returning an error?


crunk1:
See https://play.golang.com/p/YgPwRAtSQdb

```go
package main

import (
	"fmt"
	
	"github.com/samber/lo"
)

func main() {
	ints := []int{0,1,2}
	fmt.Println(lo.Nth(ints, -1))  // prints 2, as expected
	
	ints = []int{0}
	fmt.Println(lo.Nth(ints, -1))  // returns error "nth: -1 out of slice bounds" - but shouldn't
}
```

Additionally, I think this function should panic instead of raising an error when out of bounds (but I'm open to be convinced otherwise):
- the original lodash implementation doesn't return an error (returns undefined when out of bounds)
- normal Go slice accesses panic when out of bounds

I'll submit a PR with a fix - and my proposed change, but wanted to file this issue first.

PR #137, opened 2022-05-18
Nth: addressing #136 - fixing -1 bug on slice of length 1. 

crunk1:
- Fixed incorrect Nth behavior for -1 index.
- More performant index checking (haven't tested, but removed Math.Abs call and casting to/from float64)

~~Used constraints.Integer for 'nth' parameter.~~
~~Panic instead of raising error on out of range.~~

### DIFF
--- a/find.go
+++ b/find.go
@@ -2,7 +2,6 @@ package lo
 
 import (
 	"fmt"
-	"math"
 	"math/rand"
 
 	"golang.org/x/exp/constraints"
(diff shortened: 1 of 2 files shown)


crunk1:
Addressing #136

crunk1:
@dolmen removed unrelated changes

samber:
Oops.



Thanks for the fix.

PR #142, opened 2022-05-21
Add function Without which return slice with slice all elements not i…

chensylz:
…n another


samber:
Hi @chensylz 👋

Thanks for this first contribution ;)

I made 2 changes:
- collection of items as first argument and items exclude in second argument
- items to exclude as variadic argument, instead of slice

Let's merge!

PR #151, opened 2022-06-03
Nth: use constraints.Integer for nth param.

crunk1:
Mentioned in my original issue #136 and is superset to PR #137 

~~**Panic**~~
~~[Three months ago, Nth was changed to return an error instead of panicking.](https://github.com/samber/lo/commit/a59137782c3457b6039f8654b1cc9fa5af48e948) Admittedly, the explicit error handling is arguably more Go-y, however, I propose going back to panicking because:~~
~~- it's a simpler method signature, no error handling~~
~~- it is a similar behavior to normal slice accessing: `mySlice[outOfBoundsIndex]` panics~~

**constraints.Integer**
This is a more generic, backwards-compatible method signature change. Any integer type will work for `nth` parameter.

### DIFF
--- a/find.go
+++ b/find.go
@@ -208,17 +208,18 @@ func Last[T any](collection []T) (T, error) {
 
 // Nth returns the element at index `nth` of collection. If `nth` is negative, the nth element
 // from the end is returned. An error is returned when nth is out of slice bounds.
-func Nth[T any](collection []T, nth int) (T, error) {
+func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
+  n := int(nth)
 	l := len(collection)
-	if nth >= l || -nth > l {
+	if n >= l || -n > l {
 		var t T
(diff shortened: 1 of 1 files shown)


samber:
About the error handling, you can simply call `Must`: 

```go
item := lo.Must(lo.Nth([]string{"a", "b"}, 42))
```

We can add an example into the readme.

--- 

Using `constraint.Integer` instead of `int` looks cool. I'm pretty sure we can do it in many helpers.

However, your IDE will autocomplete with `nth N`. I think we should name this type more explicitly, such as `Number`.

WDYT?

crunk1:
Would `Number` be ints and floats? If so, I'd say no. I use Goland and it handles `constraints.Integer` fine.

crunk1:
@samber I have removed the panic logic. This PR now only includes the change to using `constraints.Integer`

PR #154, opened 2022-06-05
feat: adding Compact

Syuparn:
`Compact` returns a new slice with all zero values removed (`0` for `int`, `false` for `bool`, `""` for `string` and so on). This is inspired by [Lodash `compact` method](https://lodash.com/docs/4.17.15#compact).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Supported helpers for slices:
 - Reject
 - Count
 - CountBy
+- Compact
 
 Supported helpers for maps:
 
@@ -645,6 +646,17 @@ slice := lo.ReplaceAll(in, 0, 42)
 
 slice := lo.ReplaceAll(in, -1, 42)
 // []int{0, 1, 0, 1, 2, 3, 0}
+```
+
+### Compact
+
+Returns a slice of all non-zero elements.
+
+```go
+in := []string{"", "foo", "", "bar", ""}
+
+slice := lo.Compact[string](in)
+// []string{"foo", "bar"}
 ```
 
 ### Keys
--- a/slice.go
+++ b/slice.go
@@ -406,3 +406,18 @@ func Replace[T comparable](collection []T, old T, new T, n int) []T {
 func ReplaceAll[T comparable](collection []T, old T, new T) []T {
 	return Replace(collection, old, new, -1)
 }
+
+// Compact returns a slice of all non-zero elements.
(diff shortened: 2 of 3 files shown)


DillonStreator:
How does [this](https://github.com/samber/lo/pull/154/files#diff-79f6a2aa0af049d51d15ed9bd3f5c4d511763a8ae918a0a95f062a7f712283a8R417) work with a slice of structs? Would it be necessary to use reflect DeepEqual?

Syuparn:
> How does [this](https://github.com/samber/lo/pull/154/files#diff-79f6a2aa0af049d51d15ed9bd3f5c4d511763a8ae918a0a95f062a7f712283a8R417) work with a slice of structs? Would it be necessary to use reflect DeepEqual?

`reflect.DeepEqual` is not necessary. Zero values are defined for all types.

https://go.dev/ref/spec#The_zero_value

### 1. slice of structs

If all fields of an element are zero values, `Compact` removes it.

```go
type foo struct {
	bar int
	baz string
}

lo.Compact([]foo{
	{bar: 1, baz: "a"}, // all fields are non-zero
	{bar: 0, baz: ""},  // all fields are zero values
	{bar: 2, baz: ""},  // bar is non-zero
}) // []foo{{bar: 1, baz: "a"}, {bar: 2, baz: ""}}
```

### 2. slices of pointers to structs

If an element is `nil`, `Compact` removes it.

```go
r5 := Compact([]*foo{
	&foo{bar: 1, baz: "a"},
	// nil is zero value of pointers
	nil,
	// pointer to foo{bar: 0, baz: ""}. foo{bar: 0, baz: ""} is zero value of foo, but pointer to this is not a zero value of *foo.
	&foo{bar: 0, baz: ""},
	&foo{bar: 2, baz: ""},
}) // []*foo{&foo{bar: 1, baz: "a"}, &foo{bar: 0, baz: ""}, &foo{bar: 2, baz: ""}}
```

`lo.Empty` is an easy way to obtain zero values.

https://github.com/samber/lo/blob/master/type_manipulation.go#L42

DillonStreator:
@Syuparn could be worth adding those struct examples to tests

Syuparn:
@DillonStreator I added these samples to test cases.

samber:
Hi there. Thanks for the contrib.

It looks good.

I will add an `IsEmpty` helper in another PR.

PR #155, opened 2022-06-06
add Interleave helper to slices

dillonstreator:
`Interleave` interleaves input slices into single output result


samber:
nice helper!

thanks for this first contribution ;)

PR #156, opened 2022-06-07
feat: add parallel.Find

shekhirin:
Useful when you don't care about the order and need any matching element in collection ASAP.

For example, this snippet
```golang
func main() {
	start := time.Now()
	
	lo.Find([]string{"a", "b", "c", "d"}, func(letter string) bool {
		time.Sleep(1 * time.Second)
		
		return letter == "c"
	})
	
	fmt.Printf("finished Find in %s\n", time.Since(start).String())
}
```
finishes in ~3 seconds, while parallel one
```golang
func main() {
	start := time.Now()
	
	lop.Find([]string{"a", "b", "c", "d"}, func(letter string) bool {
		time.Sleep(1 * time.Second)
		
		return letter == "c"
	})
	
	fmt.Printf("finished Find in %s\n", time.Since(start).String())
}
```
in ~1 second.

### DIFF
--- a/parallel/slice.go
+++ b/parallel/slice.go
@@ -3,7 +3,7 @@ package parallel
 import "sync"
 
 // Map manipulates a slice and transforms it to a slice of another type.
-// `iteratee` is call in parallel. Result keep the same order.
+// `iteratee` is called in parallel. Result keep the same order.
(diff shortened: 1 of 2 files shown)


FournyP:
Any news ?

PR #157, opened 2022-06-08
feat: add MapToSlice method

sleagon:
Add a method to transform map[x]y to slice according to given func.

somthing like :

    	result1 := MapToSlice(map[int]int{1: 1, 2: 2, 3: 3, 4: 4}, func(x int, v int) string {
		return fmt.Sprintf("%d_%d", x, v)
        })

### DIFF
--- a/map.go
+++ b/map.go
@@ -160,3 +160,14 @@ func MapValues[K comparable, V any, R any](in map[K]V, iteratee func(V, K) R) ma
 
 	return result
 }
+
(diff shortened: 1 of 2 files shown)


shekhirin:
> By the way, I just add another two method to find the first or last matched index of the slice.

I think it's the same as https://github.com/samber/lo#findindexof?

sleagon:
> LastMatched
yes, i'll remove the second commit.

PR #164, opened 2022-06-22
add AssignBy

yoshidan:
I need the function to customize map merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,7 +791,7 @@ Merges multiple maps from left to right.
 ```go
 mergedMaps := lo.AssignBy[string, int](
     func(v1, v2) int {
-	    return v1 + v2	
+        return v1 + v2
     },
     map[string]int{"a": 1, "b": 2},
     map[string]int{"b": 3, "c": 4},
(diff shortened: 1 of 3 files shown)


samber:
Hi @yoshidan and thanks for this first contribution.

I wonder if it would be better to use a variadic callback such as `iteratee(args ...T) T`, instead of a simple `iteratee(a T, b T) T`.

I suppose in most cases, it is easier to build a callback with a fixed number of arguments, but sometimes, having all intermediary values would be very useful.

WDYT?

yoshidan:
Hi @samber and thanks for your review.

I implemented it with reference to [mergeWith](https://github.com/lodash/lodash/blob/master/mergeWith.js). 
Since `customzer` needs `oldValue` and` srcValue`, I thought it would be better to limit the argument of `iteratee` to two.
If it should be general purpose, it would be better to include the key and map itself in the callback like the [internal implementation of mergeWith](https://github.com/lodash/lodash/blob/master/.internal/baseMerge.js#L30).
```
iteratee func(V, V, K, map[K]V, map[K]V)
```
```
result1 := AssignBy(func(a int, b int, _ string, _ map[string]int, _ map[string]int) int {
     return a + b
}, map[string]int{"a": 1, "b": 2}, map[string]int{"b": 3, "c": 4})
```

I couldn't think of a use case that I wanted to have variadic arguments, so could you please tell me when it's needed?

PR #165, opened 2022-06-22
feat: add `FromPtr` function

frederik-suerig:
Hey,
wanted to add a basic function that returns the value of a pointer.

When the pointer is `nil` it returns the default value of the type. I.e. for a `nil` string pointer, it would return an empty string `""`.


samber:
Hi @frederik-suerig and thanks for this first contrib.

I gonna add some doc and test after merge.

PR #180, opened 2022-07-13
Add Remove to remove an item from a slice

MarkRosemaker:
This PR adds a function to remove an element from a slice.

h/t to https://yourbasic.org/golang/delete-element-slice/


samber:
Hi @MarkRosemaker and thanks for your first contribution.

I would suggest `RemoveIndex` instead of `Remove`, because we may decline this helper in multiple ways: `RemoveBy`, `RemoveByKey`, `RemoveByValue`...

Could you please return a new slice instead of modifying the collection ?

PR #181, opened 2022-07-13
Allow Union to take many lists

frankywahl:
Allow a union of any number of lists.

Note: similar code could be written for a number of functions, but I wanted to open this up for discussion. 

I also thought about changing the API to for more flexibility: 

```go
func Union[T comparable](lists ...[]T) []T 
```

This would be backward compatible, but would change the required number of arguments, so left it as it was for now. Happy to change it.

### DIFF
--- a/intersect.go
+++ b/intersect.go
@@ -143,7 +143,12 @@ func Difference[T comparable](list1 []T, list2 []T) ([]T, []T) {
 
 // Union returns all distinct elements from both collections.
 // result returns will not change the order of elements relatively.
-func Union[T comparable](list1 []T, list2 []T) []T {
+func Union[T comparable](list1 []T, list2 []T, lists ...[]T) []T {
+	if len(lists) > 0 {
+		l := Union(list2, lists[0])
+		return Union(list1, l, lists[1:]...)
+	}
+
 	result := []T{}
 
 	seen := map[T]struct{}{}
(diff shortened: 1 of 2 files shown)


frankywahl:
@samber Any updates on this, I've updated it recently given that `Union` has been implemented since the PR was open.

intersect.go
```diff
@@ -98,23 +98,31 @@ func NoneBy[T any](collection []T, predicate func(item T) bool) bool {
...
+	for _, elem := range lists[len(lists)-1] {
+		if _, ok := seen[elem]; ok {
+			result = append(result, elem)
 		}
 	}
 
-	return result
+	lists[0] = result
+
+	return Intersect(lists[:len(lists)-1]...)
```
samber: please don't perform recursion since it will require extra computation

```go
for i := 1; i < len(lists); i++ {
    for j := 0; j < len(lists[i]); j++ {
		if _, ok := seen[lists[i][j]]; !ok {
			delete(seen, lists[i][j])
		}
	}
    if len(seen) == 0 {
        return Slice{}
    }
}

return Keys(seen)
```

samber:
Please add an example in lo_example.go and update documentation in README.md and docs/

NathanBaulch:
This breaks CI due to a minor linting issue @samber @frankywahl. Do CI failures not block PR merges?
I only mention this because it causes all new PRs to fail CI.

frankywahl:
> This breaks CI due to a minor linting issue @samber @frankywahl. Do CI failures not block PR merges? I only mention this because it causes all new PRs to fail CI.

Sorry about this, fixed it in https://github.com/samber/lo/pull/731

samber:
@frankywahl i usually don't block PRs with linting issues to avoid round-trip and delays for merge. And sometimes, I forget to apply the fix on the master branch 😅

PR #185, opened 2022-07-20
feat: Add Associate function

qianwj:
Associate returns a map containing key-value pairs provided by transform function applied to elements of the given slice.

If any of two pairs would have the same key the last one gets added to the map.

The returned map preserves the entry iteration order of the original array.


samber:
Hi @qianwj and thanks for this first contribution.

Let's merge!

PR #188, opened 2022-07-28
Implement ChunkString

### LINKED ISSUE #187: Chunk string
It would be great to have the equivalent to `lo.Chunk` for strings.

we could have something like:
```go
strings := lo.ChunkString("123456", 2)
// []string{"12", "34", "56"}
```

So far to do that one has to implement ChunkString himself.
Another ways to do that using `lo` package is something like:
```go
strings := lo.Map(
    lo.Chunk(bytes.Runes([]byte("123456")), 2),
    func(r []rune, _ int) string { return string(r) })
```
But that's much less understandable and not very efficient.

CorentinClabaut:
Follow the best implementation found on StackOverflow by Igor Mikushkin https://stackoverflow.com/questions/25686109/split-string-by-length-in-golang

closes #187

### DIFF
--- a/string.go
+++ b/string.go
@@ -28,20 +28,20 @@ func Substring[T ~string](str T, offset int, length uint) T {
 
 // ChunkString returns an array of strings split into groups the length of size. If array can't be split evenly,
 // the final chunk will be the remaining elements.
-func ChunkString(str string, size int) []string {
+func ChunkString[T ~string](str T, size int) []T {
 	if size <= 0 {
 		panic("lo.ChunkString: Size parameter must be greater than 0")
 	}
 
 	if len(str) == 0 {
-		return []string{""}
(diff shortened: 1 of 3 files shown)


string.go
```diff
@@ -24,6 +26,36 @@ func Substring[T ~string](str T, offset int, length uint) T {
 	return str[offset : offset+int(length)]
 }
 
+// ChunkString returns an array of strings split into groups the length of size. If array can't be split evenly,
+// the final chunk will be the remaining elements.
+func ChunkString(str string, size int) []string {
```
samber, line 31: I would use the following prototype instead:

```go
func ChunkString[T ~string](str T, size int) []T
```

CorentinClabaut: Makes sense

PR #190, opened 2022-07-29
Typo in README.md MapValues should be MapToSlice

CorentinClabaut:


CorentinClabaut:
@samber `MapToSlice` iteratee is `func(V, K)` where `func(K, V)` seems more intuitive.

Would it be worth fixing that eventhough that would break existing code that uses it?

samber:
Oh yes! 

I didn't spot this error last Sunday.

This helper was added very recently. I think we can fix it today without breaking too much code.

PR #194, opened 2022-07-31
Adding SliceToMap2->SliceToMap9 + Tuples2ToMap->Tuples9ToMap

samber:
@CorentinClabaut Can i ask for a quick review please ?

I don't really like the naming of these helpers. Do you have a better idea?

I think we can also write the following helpers:
- `lo.MapToTuplesX`
- `lo.MapXToSlice`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 @samber: I sometimes forget to update this file. Ping me on [Twitter](https://twitter.com/samuelberthe) or open an issue in case of error. We need to keep a clear changelog for easier lib upgrade.
 
+## 1.28.0 (2022-08-xx)
+
+Adding:
+
+- lo.SliceToMap2 -> lo.SliceToMap9
+- lo.Tuples2ToMap -> lo.Tuples9ToMap
+
 ## 1.27.0 (2022-07-29)
 
 Breaking:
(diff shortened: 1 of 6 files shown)


samber:
GroupBy2->GroupBy9 could be nice as well.

Maybe too much complexity?

```go
numbers := []int{
    0, 1, 2,
    3, 4, 5,
    6, 7, 8,
}

groups := lo.GroupBy2[int, int, int](numbers, func(i int) (int, int) {
    return i%3, i%2
})

map[int]map[int][]int{
    0: {
        0: []int{ 0, 6 },
        1: []int{ 3 },
    },
    1: {
        0: []int{ 4 },
        1: []int{ 1, 7 },
    },
    2: {
        0: []int{ 2 },
        1: []int{ 5, 8 },
    },
}
```

CorentinClabaut (commented):
@samber, sure, no problem.

I'm not sure to get in which situation we would need these nested maps, do you have an example in mind?

I guess having a nested `map[string]map[string]Foo` Could be useful to be able to simply get the correct Foo based on 2 keys, but wouldn't that be the same as having a map `map[lo.Tuple2[string, string]]Foo`?
In one case we would access it like that: `f := m1[key1][key2]`
And in the other case we would access it like that: `f := m2[lo.T2(key1, key2)]`
The second option is slightly more verbose, but should be simpler to use.
For example, with the second option we can easily build any map with `SliceToMap` and we wouldn't need `SliceToMapX` nor `TuplesXToMap`.
Using a tuple as the key might also make it simpler to use the other map helper functions.

What do you think?

Or am I missing a use case where the nested map would be better?

PR #200, opened 2022-08-12
Add CountValues function

orccn:
Add CountValues function


samber (approved):

samber:
Thanks for the contrib.

Let's merge ;)

Issue #202, opened 2022-08-17
Feature: Add an EmtpyOr func 


kebe7jun:
`func EmptyOr[T comparable](v T, fallback T) T` returns fallback value if first argument is a zero value.
This would be useful to set the default value when implementing some interfaces.

PR #203, opened 2022-08-17
feat: add EmptyOr func

kebe7jun:
Fix https://github.com/samber/lo/issues/202


samber:
I think in that case we can just use `lo.Coalesce("", "hello world")`

Strernd:
```
	payload := SomePayload{
		SomeAttr:     lo.Coalesce[string](input.Some, input.Alternative),
	}
```
In such a case this is unhandy. Migrating from a `lo.FromPtrOr` and there is no equivalent if the type is no pointer.

PR #205, opened 2022-08-17
feat: relax FindKeyBy constraint

### LINKED ISSUE #204: `FindKeyBy` value constraint should be relaxed to `any`
Real use case: map of maps. When map's value type is another map it is not `comparable`, but this isn't required to run predicate function.

Example:

```go
var m map[K1]map[K2]V2
_, found := lo.FindKeyBy(m, func(k1 K1, v1 map[K2]V2) bool {
	_, found := lo.FindKeyBy(v1, func(k2 K2, v2 V2) bool {
		return match(v2)
	})
	return found
})
```

powerman:
Fixes #204

### DIFF
--- a/find.go
+++ b/find.go
@@ -98,7 +98,7 @@ func FindKey[K comparable, V comparable](object map[K]V, value V) (K, bool) {
 }
 
 // FindKeyBy returns the key of the first element predicate returns truthy for.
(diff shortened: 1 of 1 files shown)


samber:
Gosh! I suppose It was a copy from `lo.FindKey` that needs a `comparable` type.

Thanks for the bug report!

Issue #192, opened 2022-07-29
Uniq (and others) turn nil slice to non-nil empty slice


powerman:
This is surprising and not documented behaviour.
I believe it's the best to fix it (return nil), or otherwise at least document it.

Alucardik:
Strongly agree, this behaviour can cause strange side effects such as returning empty arrays in json fields marked as `omitempty`, instead of skipping them.

I think it is worth linking this issue to v2 roadmap – #578 , since it is a desired change.

PR #209, opened 2022-09-01
`Uniq` returns nil slice for nil slice

k3forx:
fix: https://github.com/samber/lo/issues/192

If this PR is merged, I will fix other functions 👍


slice_test.go
```diff
@@ -151,9 +151,15 @@ func TestUniq(t *testing.T) {
 	is := assert.New(t)
 
 	result1 := Uniq([]int{1, 2, 2, 1})
+	var nilSlice []int
```
shivamrazorpay, line 154: how this is sending a nil value ?? I think we are creating an empty int slice.

PR #89, opened 2022-04-08
feat: add IfF、ElseIfF、ElseF

xiaosongfu:
this PR is inspired by the issue #65 

we want If/ElseIf/Else ’s result parameter could be a function, so i add IfF/ElseIfF/ElseF functions and make this PR.

Issue #65, opened 2022-03-23
If/ElseIf/Else ’s result parameter can support the function calculation？


chenjpu:

xiaosongfu:
i have impl this on PR #89

PR #214, opened 2022-09-08
feat: add TernaryF function

nicjohnson145:
A continuation of issue https://github.com/samber/lo/issues/65 (and by extension PR https://github.com/samber/lo/pull/89)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1554,16 +1554,11 @@ result := lo.TernaryF[string](false, func() string { return "a" }, func() string
 // "b"
 ```
 
-Useful to avoid nil-pointer dereferencing in intializations
+Useful to avoid nil-pointer dereferencing in intializations, or avoid running unnecessary code
 
 ```go
-type S struct {
-    Val *string
-}
-
-s := S{}
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1544,6 +1544,28 @@ result := lo.Ternary[string](false, "a", "b")
...
+
+```go
+type S struct {
+    Val *string
+}
+
+s := S{}
+
+someStr := TernaryF[string](s.Val == nil, func() string { return "defaultValue" }, func() string { return *s.Val })
+```
```
CorentinClabaut, line 1567: It would also be useful to avoid running code for no reason.

One suggestion:
I think you can make the example more concise and demonstrate that no useless code will be run like that:
```go
var s *string
someStr := TernaryF[string](s == nil, func() string { return uuid.New().String() }, func() string { return *s })
```

WDYT?

CorentinClabaut (commented):
Looks quite helpful :)

README.md
nicjohnson145: That works too. I just made the example a reference to the problem I was facing that lead to the PR.  I'll make the update here shortly

samber (approved):

samber:
Thanks for this first contribution ;)

Let's merge.

PR #217, opened 2022-09-16
feat: add FromSlicePtr

mash:
Hi,
I'm a happy user of samber/lo.

How about adding FromSlicePtr ?

I had an alternative ToValues name in mind, but followed existing method names in the pull request.


mkarpusiewicz:
This would be nice to have, however it would probably make more sense to have the same behavior as `FromAnySlice` returning a bool false if not elements could be dereferenced. 

I'm curious what is that the reason we are missing `FromSlicePtr` now.

samber (approved):

PR #221, opened 2022-09-22
Add function Validate that creates an error when a condition is not met

### LINKED ISSUE #220: It would be nice to have a helper to create error based on boolean
It would be nice to have a helper function like:

```go
func Validate(ok bool, format string, a ...any) error {
	if !ok {
		return errors.New(fmt.Sprintf(format, a...))
	}
	return nil
}
```

That could then be used like:
```go
err := lo.Validate(age < 10, "%d is too young, must be at least %d", age, 10)
```

CorentinClabaut:
closes #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Concurrency helpers:
 
 Error handling:
 
+- [Validate](#validate)
 - [Must](#must)
 - [Try](#try)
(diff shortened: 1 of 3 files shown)


samber:
Simple. Efficace.

Let's merge ;)

PR #236, opened 2022-10-09
more variants of the Partial function, #235

### LINKED ISSUE #235: More variants of the Partial function
At the moment there is only one variant of the `Partial` function which can be used only on functions with 2 arguments. This is quite a limitation as very often you want to use this function on other functions with more arguments. 
I propose adding more variants of the `Partial` function like `Partial1`(as an alias for `Partial`), `Partial2`, `Partial3` etc. similar to `Zip2`, `Unpack3` etc.

dergus:
closes #235


domino14:
i like it

PR #240, opened 2022-10-11
chore: test coverage improvement

VictorAssunc:
I noticed that some functions and function aliases weren't being completely covered by the tests, so I decided to create and improve some tests.

I decided to separate by commits instead of PRs to not spam the repository.

I'm not sure if I chose the best approach in each test, so I'm willing to change and learn more :)

### DIFF
--- a/parallel/slice_test.go
+++ b/parallel/slice_test.go
@@ -3,21 +3,22 @@ package parallel
 import (
 	"sort"
 	"strconv"
+	"sync/atomic"
 	"testing"
-
+	
 	"github.com/stretchr/testify/assert"
 )
 
 func TestMap(t *testing.T) {
 	is := assert.New(t)
-
+	
 	result1 := Map([]int{1, 2, 3, 4}, func(x int, _ int) string {
 		return "Hello"
 	})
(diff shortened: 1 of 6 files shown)


parallel/slice_test.go
```diff
@@ -24,9 +24,21 @@ func TestMap(t *testing.T) {
 	is.Equal(result2, []string{"1", "2", "3", "4"})
 }
 
-func TestTimes(t *testing.T) {
+func TestForEach(t *testing.T) {
 	is := assert.New(t)
+	
+	count := 0
```
samber, line 30: please use 

```go
var counter uint64
atomic.AddUint64(&counter, 1)
```

instead

samber (changes requested):
hi @VictorAssunc and thanks for this first contrib

i made 1 comment

parallel/slice_test.go
VictorAssunc: ok, i will change it

VictorAssunc: done ♻️

samber (approved):

samber:
I removed the parallel execution when you use a custom rand seed ;)

PR #241, opened 2022-10-12
feat: add ChannelMerge

hhu-cc:
add two func ChannelCollector and BufferedChannelCollector for channel helper,
collect messages from multiple input channels into one channel.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -2,6 +2,7 @@ package lo
 
 import (
 	"math/rand"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


samber:
Hi @hhu-cc and thanks for this first contribution

I would name it `ChannelMerge` or `ChannelAggregation` instead of `ChannelCollector`, because the "collection" could be interpreted as a message source or an observable.

WDYT ? any other naming idea ?

Also, I wonder if it is a good idea to create an unbuffered helper.

hhu-cc:
Hi @samber 
ChannelMerge is a better idea,  I have already changed the func name and removed the unbuffered helper.

PR #248, opened 2022-10-15
add slice TakeWhile func

### LINKED ISSUE #244: Add TakeWhile()
https://lodash.com/docs/#takeWhile

It would be nicer if we have takeWhile() in lo.

yousuf64:
resolves https://github.com/samber/lo/issues/244


mattg-imprint:
Looks like this function was added in another PR at some point. Should be able to close this one.
https://lo.samber.dev/docs/core/slice#takewhile

PR #252, opened 2022-10-21
feat: add ChannelMergeBy

hhu-cc:
Hi @samber, 
Add func ChannelMergeBy, it can accept a iteratee func for ChannelMerge.

But the params style is not same with ChannelMerge, this is not user friendly.

The problem is ...<-chan T must be the last param, and `iteratee` func is offen the last param.

Would you please give some advice?


samber:
Closing in favor of #251 since this is equivalent to chaining ChannelMerge and ChannelMap.

PR #256, opened 2022-10-26
fix: panic when Slice() accepts a negative index

Bin-Huang:
Previous:
```go
lo.Slice(arr, -1, 10) // panic: runtime error: slice bounds out of range
```

After fixed:
```go
lo.Slice([]int{0, 1, 2, 3}, -10, -1) // []int{}
lo.Slice([]int{0, 1, 2, 3}, -100, 2) // []int{0, 1}
```

I think that compared to the native method(arr[1:10]), the meaning of lo.Slice() is to avoid panic caused by out-of-bounds as much as possible. So I fixed it.


Bin-Huang:
@samber Any suggestions much appreciated

samber:
Hi @Bin-Huang and sorry for the late answer.

Let's merge.

PR #257, opened 2022-10-26
FindAll function added

Wulfheart:
I think this could be a great addition as sometimes all elements need to be searched for.


samber:
Hi @Wulfheart and thanks for this first contribution. 🚀 

What is the point of this helper, since this is very similar to `lo.Filter(collection)` ? 🤔

Wulfheart:
Sorry. Found it too late and didn’t close it. 😅

PR #259, opened 2022-11-01
Noindex functions for slice

erpaher:
Added noindex versions of some functions(Filter, Map and etc) for slice


zhulik:
@samber is something blocking this PR from being merged? Me and probably all participants of https://github.com/samber/lo/issues/263 would be happy to start using it

quantumsheep:
@samber Hey 👋 It's been 1 year. It would really be cleaner for a lot of use-cases

PR #260, opened 2022-11-07
feat: add slice Merge function

cwb2819259:


samber:
Hi @cwb2819259 and thanks for this first contribution.

We already have a similar helper actually: `lo.Flatten()`.

cwb2819259:
Thank you @samber   I hadn't noticed it before 😄

PR #267, opened 2022-11-15
New helpers for concurrency: lo.Timeout + lo.Deadline + lo.Race

samber:
- Timeout: `func Timeout(duration time.Duration, callback func(done func())) error`
- Deadline: `func Deadline(duration time.Time, callback func(done func())) error`

Timeout and Deadline should not be used when a context is available.

```go
lo.Timeout(2*time.Second, func(done func()) {
    // very long job

   done()

   // free stuff
})
```

```go
lo.Deadline(myDate, func(done func()) {
    // very long job

   done()

   // free stuff
})
```

```go
funcA := func(done func()) {
    // very long job

   done()

   // free stuff
}
funcB := func(done func()) {
    // very long job

   done()

   // free stuff
}
funcC := func(done func()) {
    // very long job

   done()

   // free stuff
}

lo.Race(funcA, funcB, funcC)
```

I'm not really sure `Timeout` and `Deadline` should be added to this library, since `context.WithTimeout` and `context.WithDeadline` are highly recommended.

For all helpers, `isDone func() bool` might be passed to callbacks.

Should we add a timeout/deadline to `lo.Race()`? More helpers `lo.RaceWithTimeout()` `lo.RaceWithDeadline()`?  

Should `lo.Race()` return anything ?

WDYT ?


concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
 	}()
 	return ch
 }
+
+// Timeout returns an error if `callback` runs longer than `duration`.
+// A return in the callback is equivalent to calling `done()`.
+// Warning: you whould use `context.WithTimeout` when available.
+func Timeout(duration time.Duration, callback func(done func())) error {
+	release := make(chan struct{}) // channel will be closed by garbage collector
```
zaqbez39me, line 105: Unbuffered channel here causes a goroutine leak. When the timeout fires and the select exits, the goroutine running callback is still alive and will eventually call `done()`, which blocks forever since nobody reads from release anymore. Same issue exists in Deadline (line 125) and Race (line 144).

To fix it use the following approach with buffered channel:
```go
release := make(chan struct{}, 1)
done := sync.OnceFunc(func() {
    release <- struct{}{}
})
```

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
...
+	case <-time.After(duration):
+		return context.DeadlineExceeded
+	}
+}
+
+// Deadline returns an error if `callback` runs longer than `duration`.
+// A return in the callback is equivalent to calling `done()`.
+// Warning: you whould use `context.WithDeadline` when available.
+func Deadline(t time.Time, callback func(done func())) error {
+	release := make(chan struct{}) // channel will be closed by garbage collector
```
zaqbez39me, line 125: The issue was addressed in comment above (line 105)

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
...
+		return nil
+	case <-time.After(time.Until(t)):
+		return context.DeadlineExceeded
+	}
+}
+
+// Race blocks until a callback is processed.
+// `done()` can be used to unlock race before leaving callback function.
+func Race(callbacks ...func(done func())) {
+	release := make(chan struct{}) // channel will be closed by garbage collector
```
zaqbez39me, line 144: The issue was addressed in comment above (line 105)

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
...
+// A return in the callback is equivalent to calling `done()`.
+// Warning: you whould use `context.WithTimeout` when available.
+func Timeout(duration time.Duration, callback func(done func())) error {
+	release := make(chan struct{}) // channel will be closed by garbage collector
+	done := func() { release <- struct{}{} }
+
+	go func() {
+		callback(done)
+		done()
+	}()
```
zaqbez39me, line 111: If callback calls done() internally and then the wrapper calls `done()` again, you get a double send on an unbuffered channel, which either blocks or races depending on timing. `sync.OnceFunc` from the fix above covers this too, but worth calling out explicitly. The same is applicable for two other functions below.

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
 	}()
 	return ch
 }
+
+// Timeout returns an error if `callback` runs longer than `duration`.
+// A return in the callback is equivalent to calling `done()`.
+// Warning: you whould use `context.WithTimeout` when available.
```
zaqbez39me, line 103: `whould` -> `should`

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
...
+	case <-release:
+		return nil
+	case <-time.After(duration):
+		return context.DeadlineExceeded
+	}
+}
+
+// Deadline returns an error if `callback` runs longer than `duration`.
+// A return in the callback is equivalent to calling `done()`.
+// Warning: you whould use `context.WithDeadline` when available.
```
zaqbez39me, line 123: `whould` -> `should`

concurrency.go
```diff
@@ -93,3 +97,59 @@ func Async6[A any, B any, C any, D any, E any, F any](f func() (A, B, C, D, E, F
...
+	case <-release:
+		return nil
+	case <-time.After(time.Until(t)):
+		return context.DeadlineExceeded
+	}
+}
+
+// Race blocks until a callback is processed.
+// `done()` can be used to unlock race before leaving callback function.
+func Race(callbacks ...func(done func())) {
```
zaqbez39me, line 143: We block forever on `<-release` in case of zero callbacks.
Add guard:
```go
if len(callbacks) == 0 {
    return
}
```

concurrency_test.go
```diff
@@ -212,3 +214,68 @@ func TestAsyncX(t *testing.T) {
 		}
 	}
 }
+
+func TestTimeout(t *testing.T) {
+	t.Parallel()
+	testWithTimeout(t, 100*time.Millisecond)
+	is := assert.New(t)
+
+	err := Timeout(10*time.Millisecond, func(done func()) {
```
zaqbez39me, line 223: No test for calling `done()` twice inside the callback.
Would be good to add a case that verifies it doesn't panic or deadlock.

zaqbez39me:
Also saw your questions @samber and here's my take.

**On `Timeout` and `Deadline`**

Agree with the hesitation. If `context.WithTimeout` and `context.WithDeadline` are the recommended path, shipping these in `lo` mostly adds confusion about when to use which. I'd drop both.

**On `isDone func() bool`**

Good call. Right now callers have no way to check if they've already signalled, which makes `done()` easy to double-call by mistake. Passing `isDone` alongside `done` makes the API safer without much added complexity.

**On `lo.Race()` return value**

It should return the index of the winning callback, otherwise callers have no way to know which one finished first, which is often the whole point. Something like:

```go
func Race(callbacks ...func(done func())) int
```

**On `lo.RaceWithTimeout` / `lo.RaceWithDeadline`**

If `Timeout` and `Deadline` are dropped as standalones, having them as `Race` variants makes more sense. `Race` is the novel primitive here and timeout/deadline feel like natural extensions of it specifically. So yes, I'd lean toward `lo.RaceWithTimeout` and `lo.RaceWithDeadline` over keeping the standalone helpers.

Overall I'd ship `Race` (with return value) now and iterate on the timeout/deadline variants in a follow-up once the base API feels right.

PR #268, opened 2022-11-15
Adding lo.SafeClose(channel)

samber:
Protect against double-close panic.

```go
ch := make(chan int)

closed := SafeClose(ch)
// true

closed = SafeClose(ch)
// false
```

Equivalent to `lo.Try0(func() { close(ch) })`


samber:
Rename to CloseOnce ?

PR #269, opened 2022-11-15
Adding lo.WaitFor

samber:
```go
ok := WaitFor(func(i int) bool {
	return i % 3 == 2
}, 10*time.Millisecond, time.Millisecond)
```

### DIFF
--- a/concurrency.go
+++ b/concurrency.go
@@ -1,6 +1,9 @@
 package lo
 
-import "sync"
+import (
+	"sync"
+	"time"
+)
 
 type synchronize struct {
 	locker sync.Locker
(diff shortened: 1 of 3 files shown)


concurrency.go
```diff
@@ -93,3 +96,35 @@ func Async6[A, B, C, D, E, F any](f func() (A, B, C, D, E, F)) <-chan Tuple6[A,
 	}()
 	return ch
 }
+
+// WaitFor runs periodically until a condition is validated.
+func WaitFor(condition func(i int) bool, maxDuration time.Duration, tick time.Duration) (int, time.Duration, bool) {
```
ccoVeille, line 101: What about this


```suggestion
func WaitFor(condition func(i int) bool, maxDuration time.Duration, tick time.Duration) (int, time.Duration, bool) {
       return WaitForWithCtx(context.Background(), condition, maxDuration, tick)
}


// WaitForWithCtx runs periodically until a condition is validated or context is cancelled
 func WaitForWithCtx(ctx context.Context, condition func(i int) bool, maxDuration time.Duration, tick time.Duration) (int, time.Duration, bool) { 
         if condition(0) { 
                 return 1, 0, true 
         } 
  
         start := time.Now() 
  
         timer := time.NewTimer(maxDuration) 
         ticker := time.NewTicker(tick) 
  
         defer func() { 
                 timer.Stop() 
                 ticker.Stop() 
         }() 
  
         i := 1 
  
         for { 
                 select { 
                 case <-ctx.Done():
                         return i, time.Since(start), false
                 case <-timer.C: 
                         return i, time.Since(start), false 
                 case <-ticker.C: 
                         if condition(i) { 
                                 return i + 1, time.Since(start), true 
                         } 
  
                         i++ 
                 } 
         } 
 }
```

ccoVeille: @samber

samber: Please open a new PR.

ccoVeille: I opened
-  #480

Issue #148, opened 2022-06-02
Binary search?


zhulik:
It would be amazing to have a binary search function for arrays of camparables in the package.

WDYT?

Would you accept a PR with it?

MohitVachhani:
Hey @zhulik raised a PR regarding this!
Please pass any suggestions!

PR #277, opened 2022-11-29
feat: adding binary search function

MohitVachhani:


find.go
```diff
@@ -366,3 +366,30 @@ func Samples[T any](collection []T, count int) []T {
 
 	return results
 }
+
+// Binary_Search search the value of a collection.
+func Binary_Search(collection []int, value int) int {
```
ns-roxer, line 371: Hi! 
Unfortunately, this is not a generic method.
In order to fit the idea of issue #148 method signature should be something like that:
`func Search[T constraints.Ordered](collection []T, value T) int {}`

Issue #286, opened 2022-12-25
`Substring` can't handle utf8 characters


trim21:
```golang
package main

import (
	"fmt"

	"github.com/samber/lo"
)

func main() {
	result1 := lo.Substring("你好，世界", 0, 2)

	fmt.Printf("%s\n", result1)
}
```

https://go.dev/play/p/hvwkx8BQxPN

when processing string, it's expected to do something like this: `string(lo.Subset([]rune(s), start, end))`

trim21:
@samber new release please?

samber:
The v1.38.0 release failed. I just made a new version: v1.38.1.

PR #288, opened 2022-12-25
fix: utf8 input `Substring`

trim21:
fix https://github.com/samber/lo/issues/286

potential breaking change, because previously `lo.Substring` didn't handle utf8 input correctly.

`ChunkString` works because  `for i := string(...)` behave like `for i := []rune(string(...))`,  but we didn't convert `~string` to `[]rune` in `Substring`, `string(...)[start:end]` doesn't handle utf8.


samber:
Fixed in #327.

I added tests for the helper. Thanks for the feedback.

PR #295, opened 2023-01-03
feat : adding ValueOr

alexandreLamarre:
Motivation : reducing boilerplate when setting sensible defaults when a key is not found in a map

```go
func ValueOr[K comparable, V any](in map[K]V, key K, fallback V) V 
```

<hr/>

```go
r1 := ValueOr(map[string]int{"foo": 1}, "bar", 2)
is.Equal(r1, 2)
```


samber:
Very useful! Thanks for the contrib.

Issue #306, opened 2023-02-03
Make a variant of Assign to do a deep maps merge


mih-kopylov:
Make
```
package main

import (
	"fmt"

	"github.com/samber/lo"
)

func main() {
	result := lo.AssignDeeply(
		map[string]any{"a": 2, "b": map[string]any{"c": 1, "d": 1}},
		map[string]any{"a": 3, "b": map[string]any{"c": 2}},
	)

	fmt.Printf("%v", result)
}
```

to print

```
map[a:3 b:map[c:2 d:1]]
```

samber:
Thanks @mprimeaux and @mih-kopylov for the suggestion.

Why not use [mergo](https://github.com/imdario/mergo) instead? The promise of this library is to extend the Go stdlib with type safety. The deep merge of structs or maps is either unsafe or requires reflection.

`mergo` is a nice lib, supporting many corner cases. Can you argue why we should bring such a helper to `lo`?

Also, we need more work on this PR if we do it.

mprimeaux:
Hi @samber. Thanks for taking the time to review and reply. 

Admittedly, I am not very familiar with mergo.  It seems their [merge for maps](https://github.com/imdario/mergo/blob/6d2237a59c57f46c2a9f99d1d57a71fb3efc0943/map.go#L34) of arbitrary depth and structure uses the `reflect` package, which I’d rather avoid due to its materially negative performance impact.  I’d also prefer to deal with compile-time errors rather than potential runtime errors with reflection.

To help answer your question, would you mind helping us understand (1) your goals of `lo` and (2) where you think the design goals of `lo` should “stop” and libraries such as `mergo` should “start”?

On a related note, I’m down a research path with Data Oriented Programming (DOP) and like how aligned `lo` is with the related principles. I came across `lo` as I was searching for a Go equivalent of Lodash. And here we are.

Look forward to understanding the goals of this library as related to `mergo` and `Lodash`.

natenho:
I agree that this behavior may cause surprise. I was expecting that nested maps were also merged.
Also, the name "Assign" is a bit counterintuitive.

PR #308, opened 2023-02-14
Add support for merging multiple maps of arbitrary depth.

mprimeaux:
This PR addresses issue #306 by adding support for merging multiple maps of arbitrary depth and structure using generics. 

I decided to use generics for this rather than interfaces.

```go
func Merge(maps ...map[any]interface{}) map[any]interface{} {
	out := make(map[any]interface{})
	for _, m := range maps {
		for k, w := range m {
			if v, ok := out[k]; ok {
				switch v := v.(type) {
				case map[any]interface{}:
					out[k] = Merge(v, w.(map[any]interface{}))
				default:
					out[k] = w
				}
			} else {
				out[k] = w
			}
		}
	}
	return out
}
```

However, the shape of the API felt a bit constrained so I ultimately settled on generics.

### DIFF
--- a/map.go
+++ b/map.go
@@ -165,6 +165,28 @@ func Assign[K comparable, V any](maps ...map[K]V) map[K]V {
 	return out
 }
 
+// Merge merges multiple maps of arbitrary depth and structure from left to right.
+func Merge[K comparable, V any](maps ...map[K]V) V {
+	out := map[K]V{}
+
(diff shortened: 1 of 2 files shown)


mprimeaux:
@samber Hope all is well. Would you mind reviewing when you have a moment? Thanks much in advance.

PR #311, opened 2023-02-17
Add EmptyableToPtr

### LINKED ISSUE #310: Add function `ToNonzeroPtr` that handles zero values
I suppose it would be nice to have such a function. In my use cases it's pretty often that I need a non-nil pointer to a value only if it's nonzero.

senago:
Closes #310

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ Conditional helpers:
 Type manipulation helpers:
 
 - [ToPtr](#toptr)
+- [ToNonzeroPtr](#tononzeroptr)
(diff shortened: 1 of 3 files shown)


samber:
Hi @senago and thanks for this first contribution.

In order to match other helper names, i would rename `zero` to `empty`.

senago:
Thank you for the review! Indeed, looks like here it's better to resort to `reflect` to cover more cases. Renamed the function to `EmptyableToPtr` for consistency with both `lo` and `mo`.

samber (approved):

PR #319, opened 2023-03-01
Async should use a buffered channel to prevent goroutine leaks

Jille:
All Async* functions should use a buffered channel, so the spawned goroutine can always write its value to the channel even if there is no reader. This allows the goroutine to exit, and the channel to be garbage collected.

Imagine the following code:

```
ch := lo.Aync(doStuff)
if err := somethingElse(); err != nil {
  return err
}
return <-ch
```

If somethingElse() fails, we'll never read from the channel and the Async-goroutine will wait forever for someone to read from the channel.

### DIFF
--- a/concurrency.go
+++ b/concurrency.go
@@ -27,7 +27,7 @@ func Synchronize(opt ...sync.Locker) *synchronize {
 
 // Async executes a function in a goroutine and returns the result in a channel.
 func Async[A any](f func() A) chan A {
-	ch := make(chan A)
+	ch := make(chan A, 1)
 	go func() {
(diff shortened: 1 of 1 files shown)


Jille:
@SnoozeThis https://github.com/samber/lo/issues/320

SnoozeThis:
(https://snoozeth.is/lJXV3j6MyhE) I will wait until https://github.com/samber/lo/issues/320 is closed and then add a comment.

samber:
Well spotted!

SnoozeThis:
Resolved.

Issue #299, opened 2023-01-06
Doesn't this need a PIPE function to chain ops?


effinsky:
In Ramda, for instance, we have a pipe func to enable chaining operations on collections etc. Seems quite a boon it would be to have it here in LO --

<img width="1047" alt="Screenshot 2023-01-06 at 09 51 26" src="https://user-images.githubusercontent.com/54535543/210966162-bdf9911c-6321-41bb-81ce-da442ff9a908.png">

GCrispino:
@effinsky I thought about the same thing!

I just did a draft implementation of what I thought that a Compose and a Pipe functions would look like.
You can see it here if you'd like: https://github.com/GCrispino/lo/blob/compose/func.go#L43

These functions receive just two functions as arguments, which is not exactly what happens in their counterparts in Ramda/Lodash, where you can pass a variadic number of functions. I think this couldn't be replicated in Golang in a type safe way, but something like the `PartialN` functions and create `ComposeN` and `PipeN` functions.

If this sounds interesting to the maintainer I'd be happy to create a pull request with these changes.

effinsky:
This would be nice, since at the moment things can get really cumbersome with needless intermediate variables for collections etc between outputs that should naturally flow into further ops as inputs.

GCrispino:
@effinsky FYI, went ahead and created #325 to address this 👍🏼

samber:
Thanks for the proposal.

We added support for Go iterators  in `lo` >= 1.52. I also implemented a dedicated Go package for reactive programming: https://github.com/samber/ro

Further discussion will take place there.

PR #325, opened 2023-03-12
feat: add Compose and Pipe functions

GCrispino:
Hello!

This PR adds functions Compose and Pipe (which is requested in #299), that respectively compose two functions and pipes the application of two functions.
These are common functions in utility libraries such as Ramda and Lodash.

Something worth noting is that these functions receive just two functions as arguments, which is not exactly what happens in their counterparts in Ramda/Lodash, where you can pass a variadic number of functions. I think this couldn't be replicated in Golang in a type safe way, but something like the PartialN functions and create ComposeN and PipeN functions.

A way to mitigate this is to create counterpart functions for different sizes of arguments, like Compose3/Pipe3 for composing/piping three functions, Compose4/Pipe4 for four functions, and so on.

PS: I know that the repo's README mentions that for contributing, one should talk with @samuelberthe on twitter. I tried doing so and got no response, so I decided to create this here anyway since there's an issue for something similar to this PR's contribution (#299). My apologies if this is not desired, but if this is not of the interest of the maintainers, no worries at all :)

### DIFF
--- a/func.go
+++ b/func.go
@@ -39,3 +39,9 @@ func Partial5[T1, T2, T3, T4, T5, T6, R any](f func(T1, T2, T3, T4, T5, T6) R, a
 		return f(arg1, t2, t3, t4, t5, t6)
 	}
 }
+
+func Compose[T, U, V any](f func(U) V, g func(T) U) func (T) V {
+	return func(t T) V {
+		return f(g(t))
+	}
+}
--- a/func.go
+++ b/func.go
@@ -40,8 +40,14 @@ func Partial5[T1, T2, T3, T4, T5, T6, R any](f func(T1, T2, T3, T4, T5, T6) R, a
 	}
 }
 
+// Compose returns new function that, when called, returns the result of calling g and then f with the result from g
 func Compose[T, U, V any](f func(U) V, g func(T) U) func (T) V {
 	return func(t T) V {
 		return f(g(t))
 	}
+}
+
+// Pipe returns new function that, when called, returns the result of calling f and then g with the result from f
+func Pipe[T, U, V any](f func(T) U, g func(U) V) func (T) V {
+	return Compose(g, f)
 }
--- a/func_test.go
+++ b/func_test.go
@@ -78,3 +78,19 @@ func TestPartial5(t *testing.T) {
 	is.Equal("26", f(10, 9, -3, 0, 5))
 	is.Equal("21", f(-5, 8, 7, -1, 7))
 }
+
+func TestCompose(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
+	sumBy2 := func(x int) int { return x + 2 }
+
+	mulBy3 := func(x int) int { return x * 3 }
+
+	sumBy2AndMulBy3 := Compose(mulBy3, sumBy2)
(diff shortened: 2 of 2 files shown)


effinsky:
I'd say this definitely needs to be tested with some more complex scenarios, prepared to chain more than just 2 funcs, and work with other generic funcs from the lo lib.

GCrispino:
@effinsky do you have any more complex scenarios or generic functions from the library that would be good to test in mind?

As per chaining more than just 2 functions, I believe it's just a matter of creating additional functions that do the same with more arguments, or am I wrong? I added a new commit with definitions for Compose3 and Pipe3, which do the same as Compose and Pipe but supporting 3 functions as arguments. Is it what you're thinking about?

effinsky:
I think when you're doing data ops like this you don't know exactly how many there might be, so it'd be cool to be able to chain as many as you want without swapping out the funcs. So maybe accept a varargs list of them? Then the problem is signatures, isn't it? Since the sig for a `filter` and for a `map` will be different, and with a generics impl as inflexible as it is with Go, that might be a problem. That said, there's stuff I'm surely not seeing and so maybe that's possible!

GCrispino:
> Then the problem is signatures, isn't it
@effinsky exactly, you can't accept variadic type parameters in Golang (at least not yet). That's probably why this library has functions Partial1, Partial2, Partial3, and so on, for example, instead of having a single one that would accept variadic type parameters.

effinsky:
> > Then the problem is signatures, isn't it?
> 
> @effinsky exactly, you can't accept variadic type parameters in Golang (at least not yet). That's probably why this library has functions Partial1, Partial2, Partial3, and so on, for example, instead of having a single one that would accept variadic type parameters.

We could accept a slice of funcs, I think, but then it'd be an issue of typing them so that all the sigs can fit in that general payload type. 

Actually, tbh, I think having a pipe/compose that takes 3-5 fns is already enough for the vast majority of pipelines you might want. I don't think a mile-long pipeline would even be readable, unless we're taking very small ops, in which case they just look clear and neat. 

All this said, I'm not sure if wouldn't be better to go with a 'chain' approach that lodash has -- this accepts a bunch of fns, and when u call 'value' on the return of the last one, it calls then and probably can do some nice lazy-eval optimizations. This'd be sweet, since potentially you wouldn't have to iterate over a collection EVERY TIME you call a higher order function on it.

xyluet:
https://github.com/samber/lo/pull/395

Hi all,

I created a PR similar to this feature with more flexibility of the arguments

lanathlor:
Hello,

I've added some functions on top of this MR because I needed it. I'm dropping it here in case someone is interested.

Here is my [commit](https://github.com/lanathlor/lo/commit/bab169c16f37828e2961889a492b38b4f96c98ef)

GCrispino:
Hey @lanathlor 👋🏼 

Happy to apply your commit onto my branch if there is any interest from the maintainers on merging my PR.

@samber sorry for the nudge, but I could see you added some labels to this PR. Do you have interest on taking it in?
If you haven't that's fine too :), happy to close this if that's the case

samber:
@GCrispino @effinsky 

We added support for Go iterators  in `lo` >= 1.52. I also implemented a dedicated Go package for reactive programming: https://github.com/samber/ro

Further discussion will take place there.

PR #327, opened 2023-03-15
fix: substring support non-English strings

liushuangls:
substring unable to correctly process non-English strings at present.


samber:
Thanks for the fix ;)

Issue #80, opened 2022-03-27
Create a function to convert unindexed iteratee to indexed 


FlandiaYingman:
Let's say there's an iteratee:
```go
func iteratee(a int) int {
	return a + 1
}
```
Whenever I use functions in `lo`, it requires me to write another iteratee to wrap it:
```go
slice = lo.Map(slice, func(i int, _ int) int {
	return iteratee(i)
})
```
Well, it's acceptable, but not elegant. If `lo` could provide a function like this:
```go
func WrapIndex[T any, R any](f func(T) R) func(T, int) R {
	return func(t T, _ int) R {
		return f(t)
	}
}
```
Then the function could be written in this way:
```go
slice = lo.Map(slice, WrapIndex(iteratee))
```

It's seldom we use an indexed version iteratee, most of the time we just throw away the index. 

However, providing an unindexed version for all functions would be tedious. Instead, providing a wrapper function to throw away the index should be better.

BTW, I don't know whether I used the words 'indexed' and 'unindexed' correctly. Hope you could understand. XD

OladapoAjala:
Is this feature currently being considered?

thesiti92:
this would be surprisingly useful

eplusminus:
I would say the same for predicates— that is, with methods like `Filter(slice []T, func(T, int) bool)`, I am _far_ more likely to be using them in situations where I have a `func(T) bool`; the index is almost always irrelevant, and it's extremely annoying to have to keep writing inline anonymous wrapper funcs.

GRbit:
First of all, I'm very pleased with the functionality provided by `lo` package. It's great, many kudos to the author and contributors. 

Despite the fact, that having an index argument increases the flexibility of functions and makes them truly general-purpose functions, I can share that based on my personal experience of `lo` package usage I have never ever used `index` argument. Literally, never. It's always `func(item T, _ int) R`.

Since the package is on v1.* and needs to maintain backward compatibility, it's quite hard to change it. What I suggest is releasing v2 and changing the default iterate signature from `func(item T, index int) R` to `func(item T) R` and creating additional functions like `MapI`, ForEachI` which will accept functions with iterate. I'm not sure if it needs a separate issue, @samber what do you think? Can you please share your thoughts on that?

PR #334, opened 2023-03-29
Add wrapper functions to simplify creating iteratees

fdegiuli:
These three functions (`ToIteratee`, `PartialRight`, and `Not`) allow us to turn simple functions into effective iteratees without having to explicitly wrap them in another function declaration.

Closes issue #80

### DIFF
--- a/func.go
+++ b/func.go
@@ -39,3 +39,59 @@ func Partial5[T1, T2, T3, T4, T5, T6, R any](f func(T1, T2, T3, T4, T5, T6) R, a
 		return f(arg1, t2, t3, t4, t5, t6)
 	}
 }
+
+// Partial returns new function that, when called, has its last argument set to the provided value.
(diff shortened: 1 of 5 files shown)


func.go
```diff
@@ -39,3 +39,59 @@ func Partial5[T1, T2, T3, T4, T5, T6, R any](f func(T1, T2, T3, T4, T5, T6) R, a
 		return f(arg1, t2, t3, t4, t5, t6)
 	}
 }
+
+// Partial returns new function that, when called, has its last argument set to the provided value.
```
fdegiuli, line 43:
```suggestion
// PartialRight returns new function that, when called, has its last argument set to the provided value.
```

slice.go
```diff
@@ -592,3 +592,12 @@ func IsSortedByKey[T any, K constraints.Ordered](collection []T, iteratee func(i
 
 	return true
 }
+
```
fdegiuli, line 595: This function makes sense in this file to me since it's only really useful for slices. Happy to move it though.

Also happy to change the name.

PR #371, opened 2023-08-03
Add slice Splice an element or multiple elements at index i.

wenlingang:
Inserts an element or multiple elements at index i.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Supported helpers for slices:
 - [Compact](#compact)
 - [IsSorted](#issorted)
 - [IsSortedByKey](#issortedbykey)
+- [Splice](#Splice)
 
 Supported helpers for maps:
 
@@ -979,6 +980,17 @@ slice := lo.IsSortedByKey([]string{"a", "bb", "ccc"}, func(s string) int {
 ```
 
 [[play](https://go.dev/play/p/wiG6XyBBu49)]
+
+### Splice
+
+Splice multiple elements at index i.
+
+```go
+result := lo.Splice([]string{"a", "d"}, 1, "b", "c")
+// []string{"a", "b", "c", "d"}
+```
+
+[[play](https://go.dev/play/p/8gXdkTovsIC)]
 
 ### Keys
 
--- a/README.md
+++ b/README.md
@@ -983,14 +983,14 @@ slice := lo.IsSortedByKey([]string{"a", "bb", "ccc"}, func(s string) int {
 
 ### Splice
 
-Splice multiple elements at index i.
+Insert multiple elements at index i.
 
 ```go
 result := lo.Splice([]string{"a", "d"}, 1, "b", "c")
 // []string{"a", "b", "c", "d"}
 ```
 
-[[play](https://go.dev/play/p/8gXdkTovsIC)]
+[[play](https://go.dev/play/p/G5_GhkeSUBA)]
 
 ### Keys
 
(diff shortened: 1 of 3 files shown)


1800alex (commented):
This code looks good to me, anything preventing this from being merged? Would like to make use of it

1800alex:
On second glance, something seems off

```
package slices

import (
	"testing"
)

// Insert inserts an element at index i.
// Play: https://go.dev/play/p/Eu3pk6_ODVQ
func Insert[T any](collection []T, i int, element T) []T {
	result := make([]T, len(collection)+1)
	copy(result, collection[:i])
	result[i] = element
	copy(result[i+1:], collection[i:])

	return result
}

// Insert inserts an element at index i.
func Insert2[T any](collection []T, i int, element T) []T {
	return append(collection[:i], append([]T{element}, collection[i:]...)...)
}

func BenchmarkInsertSlice(b *testing.B) {
	data := make([]int, 0)
	for i := 0; i < b.N; i++ {
		data = Insert(data, i, i)
	}
}

func BenchmarkInsertSlice2(b *testing.B) {
	data := make([]int, 0)
	for i := 0; i < b.N; i++ {
		data = Insert2(data, i, i)
	}
}
```

Running a benchmark of this seems way off from how I would expect an insert to look.

```
goos: linux
goarch: amd64
pkg: slices
cpu: AMD Ryzen 9 7900X 12-Core Processor
BenchmarkInsertSlice-24           153384            112324 ns/op
BenchmarkInsertSlice2-24        189345357                6.104 ns/op
```

wenlingang:
> On second glance, something seems off
> 
> ```
> package slices
> 
> import (
> 	"testing"
> )
> 
> // Insert inserts an element at index i.
> // Play: https://go.dev/play/p/Eu3pk6_ODVQ
> func Insert[T any](collection []T, i int, element T) []T {
> 	result := make([]T, len(collection)+1)
> 	copy(result, collection[:i])
> 	result[i] = element
> 	copy(result[i+1:], collection[i:])
> 
> 	return result
> }
> 
> // Insert inserts an element at index i.
> func Insert2[T any](collection []T, i int, element T) []T {
> 	return append(collection[:i], append([]T{element}, collection[i:]...)...)
> }
> 
> func BenchmarkInsertSlice(b *testing.B) {
> 	data := make([]int, 0)
> 	for i := 0; i < b.N; i++ {
> 		data = Insert(data, i, i)
> 	}
> }
> 
> func BenchmarkInsertSlice2(b *testing.B) {
> 	data := make([]int, 0)
> 	for i := 0; i < b.N; i++ {
> 		data = Insert2(data, i, i)
> 	}
> }
> ```
> 
> Running a benchmark of this seems way off from how I would expect an insert to look.
> 
> ```
> goos: linux
> goarch: amd64
> pkg: slices
> cpu: AMD Ryzen 9 7900X 12-Core Processor
> BenchmarkInsertSlice-24           153384            112324 ns/op
> BenchmarkInsertSlice2-24        189345357                6.104 ns/op
> ```

Thank you. After testing, it is indeed as you said, and I modified it again

samber:
Wow, the difference between both versions is impressive!

Sorry for the delay.

I like your helper, but I'm not sure calling it `Insert` is self-explanatory.

JS, PHP, C++ and Rust named it "splice".

samber:
And I think we should leverage vaargs to write a single helper instead of 2.

wenlingang:
@samber Thank you for your suggestion. I found that it can be further optimized, and I have already done it. Additionally, the name 'Insert' is borrowed from Ruby's Array#insert, though 'Splice' is also a good name.

samber (changes requested):

samber:
I made some changes:
- handle overflow
- negative index
- insert into newly allocated slice instead of `append()`

Issue #351, opened 2023-05-31
Proposal: Add Join function on slice type


kangarooxin:
can join any type not just string
```go
func Join[T any](s []T, sep string) string {
	return JoinSliceBy(s, sep, func(t T) string {
		return fmt.Sprint(t)
	})
}

func JoinBy[T any](s []T, sep string, f func(T)string) string {
	strSlice := make([]string, len(s))
	for i, v := range s {
		strSlice[i] = f(v)
	}
	return strings.Join(strSlice, sep)
}
```

PR #373, opened 2023-08-08
feat: implement lo.Join

nguyenvantuan2391996:
Converts all elements in array into a string separated by separator.

Reference Join in Lodash: https://lodash.com/docs/4.17.15#join

implement lo.Join for issue #351: https://github.com/samber/lo/issues/351

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -1,6 +1,7 @@
 package lo
 
 import (
+	"fmt"
 	"math/rand"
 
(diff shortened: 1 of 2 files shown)


lesichkovm:
Standard go library already does it: https://pkg.go.dev/strings#Join  

Best to keeo it out, or if added to use the golang function directly underneath as its more performant.

PR #376, opened 2023-08-27
Add new method Concat

jason-zhj:
Add a new method `Concat` similar to the one in lodash, it creates a concatenated slice which is independent of the original slice, unlike `append`. This comes in handy when you need to create multiple slices from one base slice

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Supported helpers for slices:
 - [CountValuesBy](#countvaluesby)
 - [Subset](#subset)
 - [Slice](#slice)
+- [Concat](#concat)
 - [Replace](#replace)
 - [ReplaceAll](#replaceall)
 - [Compact](#compact)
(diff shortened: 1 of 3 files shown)


samber:
This is exactly what `append` does. Can you explain the goal of this PR?

Thanks.

samber:
I'm going to change the prototype to `Concat[T any](collections ...[]T) []T`.

When using `append()` for chaining multiple collections, you have to do something like that: `append(append(append(append(...), ...), ...), ...)`. Such implementation might lead to multiple reallocation and copy.

If `lo.Concat` receives a list of collections, I think we can preallocate and concat everything much faster.

samber (approved):

ccoVeille:
My questio night be stupid, but how does in differs from `slices.Concat` from standard lib ?

samber:
Uh, didn't know it exists 😂

ccoVeille:
I often find things in slices package that I was looking for years, and I'm like wow.

but you are not the only one, that's why I opened https://github.com/dominikh/go-tools/issues/1552

samber:
Thanks!

https://www.youtube.com/watch?v=gAjR4_CbPpQ

PR #377, opened 2023-09-03
tableToTree helper

andrey-tushev:
A helper that transforms database-like tables
```go
{1, 0, "A"},
{2, 1, "AA"},
{3, 1, "AB"},
{4, 3, "ABA"},
{5, 1, "AC"},
{6, 0, "B"},
```
 into tree-structures
```go
[]{
    {value: "A", leafs: []{
        {value: "AA"},
        {value: "AB", leafs: []{
             {value: "ABA"},
        }},
        {value: "AC"},
    }},
    {value: "B"}
}
```
For example: table of content, shop categories or any other hierarchy

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,10 @@ Constraints:
 
 - Clonable
 
+Tree helpers:
+
+- [TableToTree](#tabletotree)
+
 ### Filter
 
 Iterates over a collection and returns an array of all the elements the predicate function returns `true` for.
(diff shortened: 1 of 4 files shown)


lesichkovm:
In most cases you would still need the ID and the Parent ID in the tree nodes. May be just the example does not. Does this retain this info?

Also in most of the similar libraries the used word is "children" not "leafs". Best to keep consistent and use "children".

Also can you provide the reverse function from a Tree structure to Flattened-Tree, This would be helpful when needing to store it back?

Issue #391, opened 2023-11-11
proposal: get child array instead of source array in UniqBy


duc2h:
I think we should have a function to get child array instead of source array in `UniqBy`, for example:

I have an array students, I want to filter students who are duplicates and get their names only. 
```
	students := []student{
		{
			Id:   1,
			Name: "student1",
		},
		{
			Id:   2,
			Name: "student2",
		},
		{
			Id:   1,
			Name: "student1",
		},
	}
```
If I use `UniqBy` it does not adapt to my case => I want another function that can help with that.

gallone2000:
I'm working on this feature

PR #392, opened 2023-11-11
add UniqItemBy

duc2h:
The issue: [391](https://github.com/samber/lo/issues/391)


samber:
I would name it UniqByMap to respect the naming of the library (2 operations: UniqBy+Map)

duc2h:
> I would name it UniqByMap to respect the naming of the library (2 operations: UniqBy+Map)

hi @samber, I have renamed the function UniqItemBy to UniqByMap for clarity.

PR #395, opened 2023-11-17
add Chain method

### LINKED ISSUE #394: feat: add chain like js lodash flow
I want to have this function [_.Flow](https://lodash.com/docs/#flow)..

We can use generic.

xyluet:
closes #394

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -1,0 +1,21 @@
+package lo
+
+// Middleware represents a function that takes an input of type T and returns an output of type T.
+// It is used as a building block for creating middleware chains.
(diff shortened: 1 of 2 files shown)


samber:
Won't do.

Unfortunately, such implem won't allow different types of inputs and outputs.

I will release a new lib in early '24 for reactive programming. 🤞

GCrispino:
Hey @samber , would you also not find https://github.com/samber/lo/pull/325 useful? If not, I'm happy to close it 👍🏼

PR #404, opened 2023-12-06
Add FilterInPlace

lennon-guan:
We need a higher performance method to complete the slice filtering operation, specifically incorporating the FilterInPlace function.


samber:
Hi @lennon-guan and thanks for your contribution.

This is a mutable helper; therefore, this is not in the spirit of this library.

I let this PR open in case we decide to create a sub-package.

samber:
Closing in favor of #577

PR #407, opened 2023-12-09
feat: circular slice selection

victorguarana:
Adds two functions to select item by index of a slice as if it were a circular linked list. 
The difference between them is in the way they treat an empty slice, `CircularSelection` returns an empty instance while `CircularSelect` panics.
I belive `CircularSelection` is better, feel free to comment too.

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -592,3 +592,36 @@ func IsSortedByKey[T any, K constraints.Ordered](collection []T, iteratee func(i
 
 	return true
 }
+
+// CircularSelection returns one element per index of a slice as if it were a circular linked list. Returns empty element if the slice is empty.
+// Play: https://go.dev/play/p/0IafnCqCVTu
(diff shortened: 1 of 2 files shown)


samber:
Instead of adding a new helper, I would update the `lo.Nth` helper, to add your `%` operation.

However, for selecting a range of items, I would not change `lo.Slice`. Instead, I would create a new helper named `lo.Circular` or `lo.Ring`, that extracts a window with offset reset.

WDYT ?

victorguarana:
But updating the `lo.Nth` operation with `%` would change its behavior. Because it would no longer return an error when the index is greater than the collection length.

Taking advantage of your idea, do you think it would be a good idea to rename the method to `lo.Circular` and move it to the `finder.go` file?

victorguarana:
@samber Changes ready. Can you check if it's ok now?

PR #409, opened 2023-12-14
feat: return nil for slice operations that received nil

abemedia:
Amend slice functions to return `nil` if the collection argument is `nil`.

This is a non-breaking change because all built-ins etc. treat nil slices the same as zero length slices.

I also fixed some of the tests as they had expected and actual in the wrong order.

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -9,6 +9,10 @@ import (
 // Filter iterates over elements of collection, returning an array of all elements predicate returns truthy for.
 // Play: https://go.dev/play/p/Apjg3WeSi7K
 func Filter[V any](collection []V, predicate func(item V, index int) bool) []V {
+	if collection == nil {
+		return nil
+	}
+
 	result := make([]V, 0, len(collection))
 
 	for i, item := range collection {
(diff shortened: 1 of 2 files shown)


samber (approved):

samber:
Can you argue about returning `nil` instead of zero value (an empty array).

Because if I try to apply the same rule to other helpers, what would be the behavior of `SumBy`:

```go
func SumBy[T any, R constraints.Float | constraints.Integer | constraints.Complex](collection []T, iteratee func(item T) R) R {
	var sum R = 0
		sum = sum + iteratee(item)
	}
	return sum
}
```

abemedia:
Hi @samber 

The behaviour of `SumBy` wouldn't change at all as iterating over a nil slice is a perfectly valid operation.

Thanks for approving! Just seen there's a bunch of conflicts so will fix ASAP.

abemedia:
Have fixed the conflicts. Regarding arguing why it should be nil, not zero, If I pass a nil slice into a function to remove elements and it was nil and then it comes back as a zero slice that feels semantically wrong. The behaviour of returning nils when that was passed in is also what the stdlib does in the `slices` package. See https://go.dev/play/p/kC2YOFdslWO

The reason I was inspired to fix this is because a load of my tests were breaking when I started using `lo` for filtering and mapping slices as our REST API was suddenly returning JSON arrays for fields which were meant to be nulls.

abemedia:
@samber any chance of getting this merged?

PR #417, opened 2024-01-06
add ToQuery and FromQuery

shayan0v0n:
The ToQuery() function is a powerful utility in Go that facilitates the effortless conversion of a map into a well-formed query string. This functionality becomes particularly valuable when you need to create query strings for web requests or other scenarios where structured key-value pairs are essential.

The ToQuery() function streamlines the process of transforming a map into a query string by intelligently handling the concatenation of key-value pairs. Let's delve into how this function simplifies the generation of query strings with a concrete example.

Consider the following scenario where you have a map representing user information:

`map[email:joan@gmail.com token:jkbcqp1wu2b user:joan]`

The desired output is a well-formatted query string:

`user=joan&email=joan@gmail.com&token=jkbcqp1wu2b`
 
Complementing the ToQuery() function, the FromQuery() function seamlessly reverses the process. It transforms a well-formed query string back into a map, providing a versatile bidirectional solution for scenarios requiring both query string creation and parsing.

`user=joan&email=joan@gmail.com&token=jkbcqp1wu2b`

ere, the FromQuery() function adeptly parses the query string, reconstructing the original map:

`map[email:joan@gmail.com token:jkbcqp1wu2b user:joan]`

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -1,0 +1,39 @@
+package lo
+
+import (
+	"fmt"
+	"strings"
+)
+
+// form query convert string to query string
+func ToQuery[K comparable, V any](in map[K]V) string {
+	res := ""
+
+	firstIteration := true
+	for k, v := range in {
+		if firstIteration {
+			res += fmt.Sprintf("%v=%v", k, v)
+			firstIteration = false
+		} else {
+			res += fmt.Sprintf("&%v=%v", k, v)
+		}
+	}
+
+	return res
+}
+
+// form query convert query string to string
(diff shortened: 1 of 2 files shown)


lesichkovm:
Several comments:

1. The Query wording is ambiguous. Most people, including me, will think about SQL Query, not about URL manipulations.

2. These functions seem to lack edge case test, like variable escaping, and I would consider to have security flaws.

3. Seem to already exist in Go: https://pkg.go.dev/net/url#ParseQuery

PR #418, opened 2024-01-09
Fix IsNil link for pkg.go.dev

invzhi:
Use lower case to make sure the link in pkg.go.dev is working.


invzhi:
@samber Could you please merge this quick fix?

samber (approved):

PR #419, opened 2024-01-11
Custom error type.

RelicOfTesla:


heww:
Hi, could someone take a look at this MR? The `MustX` function will lose the original error.

RelicOfTesla:
> Hi, could someone take a look at this MR? The `MustX` function will lose the original error.

Can you give me a example code ?

RelicOfTesla:
Haha, this is right. I misunderstood your meaning.

samber:
Can you give an example of how these helpers would be used?

I don't see in what context it could be useful. But I might be wrong.

RelicOfTesla:
> Can you give an example of how these helpers would be used?
> 
> I don't see in what context it could be useful. But I might be wrong.

You can see the "TestErrorWrapUnWrap" and "TestErrorGlobalErrHandler" func

PR #420, opened 2024-01-12
Add Identity function

l0rem1psum:
Adds an identity function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ Type manipulation helpers:
 - [IsEmpty](#isempty)
 - [IsNotEmpty](#isnotempty)
 - [Coalesce](#coalesce)
+- [Identity](#identity)
 
 Function helpers:
 
@@ -2342,6 +2343,17 @@ var nilStr *string
 str := "foobar"
 result, ok := lo.Coalesce[*string](nil, nilStr, &str)
 // &"foobar" true
+```
+
+### Identity
+
+Returns the argument itself.
+```go
+lo.Identity(42)
+// 42
+
+lo.Identity("foobar")
+// "foobar"
 ```
 
 ### Partial
--- a/type_manipulation.go
+++ b/type_manipulation.go
@@ -106,3 +106,8 @@ func Coalesce[T comparable](v ...T) (result T, ok bool) {
 
 	return
 }
+
+// Identity returns the argument itself.
+func Identity[T any](v T) T {
+	return v
+}
--- a/type_manipulation_test.go
+++ b/type_manipulation_test.go
@@ -233,3 +233,23 @@ func TestCoalesce(t *testing.T) {
 	is.Equal(result10, struct1)
 	is.True(ok10)
 }
+
+func TestIdentity(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
+	is.Equal(Identity[int](42), 42)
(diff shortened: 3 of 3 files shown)


ccoVeille:
Same question here.

l0rem1psum:
One use case I can think of now is when you want to dynamically apply a series of operations on a type. For instance:

```go
package main

import "fmt"

type Operation[T any] func(t T) T

func main() {
	value := 1

	// operations are dynamically populated
	operations := []Operation[int]{
		identity, double, triple,
	}

	for _, op := range operations {
		value = op(value)
	}

	fmt.Println(value)
}

func identity(n int) int {
	return n
}

func double(n int) int {
	return n * 2
}

func triple(n int) int {
	return n * 3
}
```

ccoVeille:
I assumed it would be something like that.

Let's wait for maintainer's feedback

lesichkovm:
May be its me, but I still cannot see a use case where this is usable. It basically returns the value without any checks or modifications.

l0rem1psum:
> May be its me, but I still cannot see a use case where this is usable. It basically returns the value without any checks or modifications.

I think you basically just explained the purpose of identity function yourself: return the value unchanged.

It's basically like the number 0 in addition and number 1 in multiplication. It's neutral but not unimportant.

I'm no type theorist but I personally think of identity as a "default" function for functions. In the above example, imagine how would you represent a no-op amongst other ops.

lesichkovm:
> I think you basically just explained the purpose of identity function yourself: return the value unchanged.
> 
> It's basically like the number 0 in addition and number 1 in multiplication. It's neutral but not unimportant.
> 
> I'm no type theorist but I personally think of identity as a "default/placeholder" function for functions. In the above example, imagine how would you represent a no-op amongst other ops.

If there is nothing to do, I would not put it in the array of opperations on the first place:

Instead of:
```
operations := []Operation[int]{
	identity, double, triple,
}
```
Just not include the noop operation:
```
operations := []Operation[int]{
    double, triple,
}
```

l0rem1psum:
> > I think you basically just explained the purpose of identity function yourself: return the value unchanged.
> 
> > 
> 
> > It's basically like the number 0 in addition and number 1 in multiplication. It's neutral but not unimportant.
> 
> > 
> 
> > I'm no type theorist but I personally think of identity as a "default/placeholder" function for functions. In the above example, imagine how would you represent a no-op amongst other ops.
> 
> 
> 
> If there is nothing to do, I would not put it in the array of opperations on the first place:
> 
> 
> 
> Instead of:
> 
> ```
> 
> operations := []Operation[int]{
> 
> 	identity, double, triple,
> 
> }
> 
> ```
> 
> Just not include the noop operation:
> 
> ```
> 
> operations := []Operation[int]{
> 
>     double, triple,
> 
> }
> 
> ```
> 
> 
> 
> 

Yes, but what if the operation comes from another function? (This is most likely the case in reality) For examples, 

``` go
func getIntOperations() Operation[T] {
    // ...
}
```

How would you determine there is no op?

lesichkovm:
@l0rem1psum I think its a case of weird (bad) logic in the application itself, rather than real need. Probably its worth revisitingthe case of why you need it in your application. If you do not need to do anything on a variable, leave it out, instead of wrapping it in an empty function.

Issue #411, opened 2023-12-27
IntersectBy,DifferenceBy,UnionBy for slice


btagrass:
for example: DifferenceBy[T any](list1, list2 []T, equal func(t1, t2 T) bool) ([]T, []T)

takanuva15:
Can we make it vararg signatures so it'll accept any number of collections?

lydamian:
+1 this it would be super helpful

PR #426, opened 2024-02-05
dll/difference-by: Added difference-by

lydamian:
#### Issue
Added difference-by

#### Summary of changes
`
// DifferenceBy returns the difference between two collections: the elements from list1 that are not present in list2,
// and the elements from list2 that are not present in list1, based on the provided predicate function.
// The predicate function is used to compare elements from both lists.
// It takes two arguments of type T and returns a boolean value indicating whether the elements are considered equal.
// The returned slices maintain the order of the elements in the original lists.
func DifferenceBy[T any](
	list1 []T,
	list2 []T,
	predicate func(T, T) bool,
) ([]T, []T)`

I will call out though that this predicate signature/implementation is different than the lodashjs implementation found here https://lodash.com/docs/4.17.15#differenceBy but is consistent with the lodashjs version of differenceWith https://lodash.com/docs/4.17.15#differenceWith

Given that the sander/lo package doesn't have a pattern of using <function-name>With(...) I think the function name is correct/consistent with pattern laid out by other functions in this Repo... (this differenceBy is consistent with the predicate signature in other <function-name>By in the intersect.go file (e.g NoneBy(...) ContainsBy(...)))

#### Related issues
https://github.com/samber/lo/issues/411
https://github.com/samber/lo/pull/289

### DIFF
--- a/intersect.go
+++ b/intersect.go
@@ -141,10 +141,15 @@ func Difference[T comparable](list1 []T, list2 []T) ([]T, []T) {
 	return left, right
 }
 
+// DifferenceBy returns the difference between two collections: the elements from list1 that are not present in list2,
+// and the elements from list2 that are not present in list1, based on the provided predicate function.
+// The predicate function is used to compare elements from both lists.
+// It takes two arguments of type T and returns a boolean value indicating whether the elements are considered equal.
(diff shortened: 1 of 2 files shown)


lydamian:
Ready for review @samber

Issue #396, opened 2023-11-21
Proposal: Add throttle


adrianrudnik:
Based on [lodash _.throttle](https://lodash.com/docs/#throttle).

Use case: I want to publish an update to a websocket at most any given `time.Duration`. So basically like the already existing [debounce](https://github.com/samber/lo#debounce), but limiting with delay, not delaying until end.

Working with onboard functions, I came up with something like this:

```
delay := 500 * time.Millisecond

internalInvoke, cancel = lo.NewDebounceBy(delay, func(key int64, count int) {
    // .... do stuff
})

throttle = func() {
    internalInvoke(time.Now().Round(delay).UnixNano())
}
```

So in the end, calling `throttle()` will put the rounded nano time in a bucket for debounce and that bucket will be triggered after `delay` is no longer inside the same rounding range.

Not sure what to do with `cancel` though.

Lee-Minjea:
I was looking for 'throttle'

Hi @samber, if there is no PR working on this currently.
I would like to make one.

and Please pardon my absence of acknowledge on twitter, I don't have twitter account.

PR #427, opened 2024-02-14
Add NewThrottle

Lee-Minjea:
### Description

This PR adds lodash style 'throttle'.
NewThrottle function create a throttled instance that grants given functions are invoked only once in each interval.
It returns 2 functions, First one is throttled function and Second one is purge function which invokes given functions immediately and reset interval timer.

### Related Issue

#396 

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,7 @@ Concurrency helpers:
 - [AttemptWhileWithDelay](#attemptwhilewithdelay)
 - [Debounce](#debounce)
 - [DebounceBy](#debounceby)
+- [Throttle](#throttle)
 - [Synchronize](#synchronize)
 - [Async](#async)
 - [Transaction](#transaction)
@@ -3416,6 +3417,47 @@ cancel("second key")
 ```
 
 [[play](https://go.dev/play/p/d3Vpt6pxhY8)]
+
+### Throttle
+`NewThrottle` creates a throttled instance that invokes given functions only once in every interval.
+This returns 2 functions, First one is throttled function and Second one is a function to reset interval.
+
+```go
+
+f := func() {
+	println("Called once in every 100ms")
+}
+
+throttle, reset := lo.NewThrottle(100 * time.Millisecond, f)
+
+for j := 0; j < 10; j++ {
+	throttle()
+	time.Sleep(30 * time.Millisecond)
+}
+
+reset()
+throttle()
+
+```
+
+`NewThrottleWithCount` is NewThrottle with count limit, throttled function will be invoked count times in every interval.
+```go
+
+f := func() {
+	println("Called three times in every 100ms")
+}
+
+throttle, reset := lo.NewThrottle(100 * time.Millisecond, f)
+
+for j := 0; j < 10; j++ {
+	throttle()
+	time.Sleep(30 * time.Millisecond)
+}
+
+reset()
+throttle()
+
+```
 
 ### Synchronize
 
--- a/retry.go
+++ b/retry.go
@@ -287,4 +287,62 @@ func (t *Transaction[T]) Process(state T) (T, error) {
 	return state, err
 }
 
-// throttle ?
+type throttle struct {
+	mu         *sync.Mutex
+	timer      *time.Timer
+	interval   time.Duration
+	callbacks  []func()
+	countLimit int
+	count      int
+}
+
+func (th *throttle) throttledFunc() {
+	th.mu.Lock()
+	defer th.mu.Unlock()
+	if th.count < th.countLimit {
+		th.count++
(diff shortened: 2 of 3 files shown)


samber:
I think it has been implemented already: `lo.NewDebounce`.

Lee-Minjea:
> I think it has been implemented already: `lo.NewDebounce`.

Throttle and debounce work slightly differently. 
Debounce delays execution if there are new calls within a given time period, eventually it is executed only once at the end.
Throttle ensures that even if a function is called continuously, it will only be executed once within a given duration.

retry.go
```diff
@@ -287,4 +287,55 @@ func (t *Transaction[T]) Process(state T) (T, error) {
...
+	th.mu.Lock()
+	defer th.mu.Unlock()
+
+	if th.timer != nil {
+		th.timer.Stop()
+	}
+
+	if th.needInvoke || forcePurge {
+		for _, f := range th.callbacks {
+			go f()
```
samber, line 319: by default, i would make it synchronous and let the developer trigger a goroutine

WDYT ?

retry.go
```diff
@@ -287,4 +287,55 @@ func (t *Transaction[T]) Process(state T) (T, error) {
...
+	defer th.mu.Unlock()
+	th.needInvoke = true
+	if th.timer == nil {
+		th.timer = time.AfterFunc(th.interval, func() {
+			th.purge(false)
+		})
+	}
+}
+
+func (th *throttle) purge(forcePurge bool) {
```
samber, line 309: sed /purge/reset/

retry.go
samber: can you explain why you trigger the callbacks when calling purge() ?

IMO, to force a call, we would code something like this:

```go
throttle()
throttle() // skipped
throttle() // skipped
reset()
throttle()
```

retry.go
```diff
@@ -287,4 +287,55 @@ func (t *Transaction[T]) Process(state T) (T, error) {
...
+			th.purge(false)
+		})
+	} else {
+		th.timer = nil
+	}
+}
+
+// NewThrottle creates a throttled instance that invokes given functions only once in every interval.
+// This returns 2 functions, First one is throttled function and Second one is purge function which invokes given functions immediately and reset interval timer.
+func NewThrottle(interval time.Duration, f ...func()) (func(), func()) {
```
samber, line 332: I'm also think about a `NewThrottle(100ms, 3, cb)` that would trigger the first 3 calls, then ignore others during 100ms. Like we do for a basic rate limiter.

Maybe it is a different helper.

WDYT?

samber (commented):
I agree with you. This helper seems useful.

retry.go
Lee-Minjea: > by default, i would make it synchronous and let the developer trigger a goroutine
> 
> WDYT ?

Your way is more simple and configuable, 
I'll fix it that way.


> can you explain why you trigger the callbacks when calling purge() ?
> 
> IMO, to force a call, we would code something like this:
> 
> ```go
> throttle()
> throttle() // skipped
> throttle() // skipped
> reset()
> throttle()
> ```

I wanted to offer a function to force triggering callbacks and reset current interval schedule.
The usage of it is exactly same as you mentioned.
I thought, name 'reset' might imply resetting interval schedule, but not including calling the functions

retry.go
Lee-Minjea: > I'm also think about a `NewThrottle(100ms, 3, cb)` that would trigger the first 3 calls, then ignore others during 100ms. Like we do for a basic rate limiter.
> 
> Maybe it is a different helper.
> 
> WDYT?

That would be a interesting option, but I am also afraid that might makes the function's signature complicated.
So I think, it's better to be implemented in other helper or function with different signature.

samber: 2 helpers `NewThrottle` and `NewThrottleWithCount` ?

Lee-Minjea: ```NewThrottleWithCount``` looks nice,
I'll add it

retry.go
Lee-Minjea: I change the function to 'reset'.

README.md
```diff
@@ -3417,6 +3418,47 @@ cancel("second key")
...
+```
+
+`NewThrottleWithCount` is NewThrottle with count limit, throttled function will be invoked count times in every interval.
+```go
+
+f := func() {
+	println("Called three times in every 100ms")
+}
+
+throttle, reset := lo.NewThrottle(100 * time.Millisecond, f)
```
a-menshchikov, line 3450: It looks like the `NewThrottleWithCount` should be called here.

PR #429, opened 2024-02-21
style: no extra any type parameter

d-enk:
Long lines are less long

One style with

https://github.com/samber/lo/blob/5777c5a3ee09f852d55a5bb5f585fcaeb5a0aedb/errors.go#L187

https://github.com/samber/lo/blob/5777c5a3ee09f852d55a5bb5f585fcaeb5a0aedb/func.go#L37

In the doc signature in the official vscode extension are also omitted
image

I would omit it here too, how does it do vscode, but it's not necessary
https://github.com/samber/lo/blob/5777c5a3ee09f852d55a5bb5f585fcaeb5a0aedb/slice.go#L25

image

### DIFF
--- a/concurrency.go
+++ b/concurrency.go
@@ -50,7 +50,7 @@ func Async1[A any](f func() A) <-chan A {
 }
 
 // Async2 has the same behavior as Async, but returns the 2 results as a tuple inside the channel.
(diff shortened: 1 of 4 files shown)


samber:
i agree

PR #430, opened 2024-03-01
feat: add In[T]([]T, T)bool func in find package

loveRhythm1990:
add `func In[T comparable](collection []T, element T) bool ` in find package.

It is more brief and efficient than `func Find[T any](collection []T, predicate func(item T) bool) (T, bool) `


loveRhythm1990:
oh，I found this duplicated with `Contains`, close this pr

PR #434, opened 2024-03-06
feat(slice): add SplitBy function to split slices based on an iteratee

kilianpaquier:
**Purpose**

Add `SplitBy(collection []T, predicate func(item T, index int) bool) ([]T, []T)` to `slice.go` features.

A little more intuitive than `PartitionBy` when trying to split a slice in two slices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ Supported helpers for slices:
 - [Compact](#compact)
 - [IsSorted](#issorted)
 - [IsSortedByKey](#issortedbykey)
+- [SplitBy](#splitby)
 
 Supported helpers for maps:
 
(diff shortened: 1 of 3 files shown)


samber:
Some people might want to split in 3 lists, 4 lists, 5 lists...

I suppose a `PartitionBy[T, bool](...) map[bool][]T` would be simpler.

kilianpaquier:
I think that with your point of view, `GroupBy` should do the job

kilianpaquier:
`GroupBy` should do the job with this point of view !

samber:
You can use lo.FilterReject that does a similar job.

PR #439, opened 2024-04-16
feat: range-over-func POC

NathanBaulch:
I was surprised to see no mention of the upcoming range-over-func experimental feature in the issues/PR list. I haven't measured anything but in theory this feature could dramatically reduce slice allocations when composing "queries" over large collections since no intermediate slices need to be created.

Is there any interest in this? This POC PR demonstrates what iterator support might look like, including tests and examples. I went with an "I" suffix (to mirror the "F" convention) to differentiate the iterator functions from their `slice.go` counterparts.

For example:
```golang
iter := lo.ToIterator(1, 2, 3, 4)
result := lo.FilterI(iter, func(nbr int, index int) bool { return nbr%2 == 0 })
fmt.Printf("%v", result.Slice())
```

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -1,0 +1,77 @@
+//go:build goexperiment.rangefunc
+
+package lo
+
+// Iterator TODO
+type Iterator[V any] func(func(int, V) bool)
+
+// ToIterator TODO
+func ToIterator[V any](collection ...V) Iterator[V] {
+	return func(yield func(int, V) bool) {
+		for i, item := range collection {
+			yield(i, item)
+		}
+	}
+}
+
+// Len TODO
+func (iter Iterator[V]) Len() int {
+	var n int
+	for _, _ = range iter {
+		n++
+	}
+	return n
+}
+
+// Slice TODO
+func (iter Iterator[V]) Slice() []V {
+	var result []V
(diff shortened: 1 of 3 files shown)


NathanBaulch:
Probably worth thinking about if/how this would integrate with the new [iter](https://go.dev/src/iter/iter.go) package in the stdlib.

samber (commented):
I'm considering a new "iter" subpackage.

samber/lo is widely used and upgrading to Go 1.22 seems rough.

I think we can keep the same naming for every helpers, but under samber/lo/iter.

WDYT ?

iterator.go
```diff
@@ -0,0 +1,77 @@
+//go:build goexperiment.rangefunc
+
+package lo
+
+// Iterator TODO
+type Iterator[V any] func(func(int, V) bool)
+
+// ToIterator TODO
+func ToIterator[V any](collection ...V) Iterator[V] {
```
vkhobor, line 9: `ToIterator` might already be implemented in the go standard library as `slices.All`. 

As per this comment: https://github.com/gomoni/it/issues/12#issuecomment-2234202444

NathanBaulch:
Implemented in #672

Issue #390, opened 2023-11-10
Add Same function


baozaolaoba-top:
```go
func Same[T comparable](list1 []T, list2 []T) bool{
  left,right := Difference(list1,list2)
  return len(left)==0 && len(right)==0
} 
```

MohitVachhani:
The above mentioned PR might solve this out.

PR #440, opened 2024-04-17
feat: implementing the same function [Issue - 390]

MohitVachhani:
Solves:
* https://github.com/samber/lo/issues/390


matfax:
Isn't [google/go-cmp](https://pkg.go.dev/github.com/google/go-cmp) a better scope for such functionality, and already exist there with excellent computational complexity?

PR #442, opened 2024-04-17
Add FromSlicePtr

KenxinKun:
This proposal implements the reverse of `ToSlicePtr` by using `FilterMap` internally to skip any `nil` pointer values in an input array of pointers and returning the corresponding array of values safely.

### DIFF
--- a/type_manipulation.go
+++ b/type_manipulation.go
@@ -43,6 +43,16 @@ func FromPtrOr[T any](x *T, fallback T) T {
 	return *x
 }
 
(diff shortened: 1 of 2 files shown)


samber (approved):

samber (changes requested):
I would use the same behavior of `lo.FromPtr`, that returns a zero value when a pointer is nil.

It can be combined with `lo.Compact`.

I suppose you can implement a `lo.FromSlicePtrOr` helper as well.

samber:
Closing in favor of #217

PR #443, opened 2024-04-18
Implemented GroupByMap

maryum375:
Implemented a new function to GroupBy.
This Implementation allows to map the values on the same iteration.
Sending an identity function as the `iterateeValue` will de facto result in a normal GroupBy operation.

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -188,6 +188,21 @@ func GroupBy[T any, U comparable, Slice ~[]T](collection Slice, iteratee func(it
 	return result
 }
 
+
+
+// GroupByMapValues returns an object composed of keys generated from the results of running each element of collection through iterateeKey and values running each element through iterateeValue.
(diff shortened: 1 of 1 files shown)


slice.go
```diff
@@ -161,6 +161,21 @@ func GroupBy[T any, U comparable](collection []T, iteratee func(item T) U) map[U
...
 }
 
+
+
+// GroupByMapValues returns an object composed of keys generated from the results of running each element of collection through iterateeKey and values running each element through iterateeValue.
+func GroupByMapValues[K comparable, T any, V any](arr []T, iterateeKey func(T) K, iterateeValue func(T) V) map[K][]V {
+	result := map[K][]V{}
+
+	for _, item := range arr {
+		k := iterateeKey(item)
```
samber, line 171: I would use a single iteratee: `k, v := iteratee(item)`

samber (commented):
i think we need to think a little bit about the naming

your variant: `map` then `group by`

other cool variant I use in my own projects: `map` then `group by` then `aggregate values`

PR #450, opened 2024-05-11
Add DoInAttemptWithDelay and DoInAttempt functions

marktimarev:
The `Attempt` like functions are awesome and help a lot with operations prone to fail, i.e. networking stuff.

Nevertheless, only procedure implementation is available in the library. So, it forced me to use constructions like the following:
```go
fooOut, err := func() (FooOut, error) {
	var fooOut FooOut
	var err error
	_, _ = Attempt(10, func(_ int) error {
		t, err = f()
		return err
	})
	return fooOut, err
}()
```

I ended up doing this stuff so often, so I decided to contribute if you do not mind :)

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -288,3 +288,75 @@ func (t *Transaction[T]) Process(state T) (T, error) {
 }
 
 // throttle ?
+
+// DoInAttemptWithDelay invokes a function N times with the delay between calls until it returns without error.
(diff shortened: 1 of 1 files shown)


samber:
Why not just implement Attempt0, Attempt1, Attempt2 ... AttemptX ?

marktimarev:
> Why not just implement Attempt0, Attempt1, Attempt2 ... AttemptX ?

it's been a while since your reply 😄 

Actually, you are completely right, I will rewrite the functions to ones with numbers at the end 👌

PR #451, opened 2024-05-12
Add First and FirstOrZeroValue functions

Alireza-Kiani:
adding two new functions to `find.go` for fetching the first element of the slice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,8 @@ Supported search helpers:
 - [Latest](#latest)
 - [Last](#last)
 - [First](#first)
-- [FirstOrZeroValue](#firstorzerovalue)
+- [FirstOrEmpty](#FirstOrEmpty)
+- [FirstOr](#FirstOr)
 - [Nth](#nth)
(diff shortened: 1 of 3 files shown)


samber:
To follow the go conventions, I would return `value, ok` instead of `value, err`.

To follow the samber/lo convention, I would replace "zero" with "empty" in naming.

```gp
func First[T]([]T) T

func FirstOrEmpty[T]([]T) T

func FirstOr[T]([]T, T) T
```

WDYT ?

ccoVeille:
I agree. Also, why doesn't FirstOrZero depends on First ?

Alireza-Kiani:
Hey guys, thanks for your replies. @samber You are right, the initial changes did not follow conventions properly, Updated the implementation according to your saying.

@ccoVeille Good point, I made a change to the implementation and right now other functions rely on the base `First` function for finding the first element instead of duplicate code.

samber (approved):

PR #454, opened 2024-05-18
feat(map): add GetOrSet for get value of exist key or set it

AuZhoomLee:
GetOrSet is like setdefault(key, default=None) in Python for dict , and map.getOrDefault(Object key, V defaultValue), returns the value of the given key or set the fallback value if not present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -958,6 +958,21 @@ value := lo.ValueOr[string, int](map[string]int{"foo": 1, "bar": 2}, "baz", 42)
 // 42
 ```
 
+[[play](https://go.dev/play/p/pvoEPxdiL8m)]
+
+### _GetOrSet_
+
+Returns the value of the given key or the fallback value if the key is not present.
+
+```go
+value := lo.GetOrSet[string, int](map[string]int{"foo": 1, "bar": 2}, "foo", 42)
+// 1
+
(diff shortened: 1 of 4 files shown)


AuZhoomLee:
Mostly, I used to collect value from another, but not sure is set already, I will make sure first then to get it,

```go
...
v, ok := kv[k]
if !ok {
    kv[k] = defaults.ConstantValue
}
...
```

In particular, I read config form environment variables or config file to get info, there will always be a default value to make sure essential value has been set to further using.

lesichkovm:
Is this not the same as ValueOr ?

https://github.com/samber/lo?tab=readme-ov-file#valueor

AuZhoomLee:
> Is this not the same as ValueOr ?
> 
> https://github.com/samber/lo?tab=readme-ov-file#valueor

@lesichkovm  not same, ValueOr doesn't effect origin,  this one can set default to origin

lesichkovm:
@AuZhoomLee tahnk you for clarifying. Two points:

1. I can see why this is useful. Its best to make it very clear we do modify the original array, as its kind of hidden. May be better name would help woth this. 

2. Probably make the name similar to the current ValueOr so that they are alpahbetically ordered next to each other, making it easier to be found by devs.

Possible options:
- ValueOrSetDefault
- ValueOrSetIfMissing

PR #465, opened 2024-06-14
ToSlicePtr: reduce allocations and improving speed

yanmhlv:
Hi, I found a way to reduce the number of allocations and improve the overall performance of `ToSlicePtr`.

here is simple benchmark:
```
name           old time/op    new time/op    delta
ToSlicePtr-10    1.18ms ± 6%    0.25ms ± 5%   -78.60%  (p=0.000 n=14+12)

name           old alloc/op   new alloc/op   delta
ToSlicePtr-10    1.60MB ± 0%    0.80MB ± 0%   -49.92%  (p=0.000 n=15+15)

name           old allocs/op  new allocs/op  delta
ToSlicePtr-10      100k ± 0%        0k ± 0%  -100.00%  (p=0.000 n=15+15)
```

### DIFF
--- a/slice_benchmark_test.go
+++ b/slice_benchmark_test.go
@@ -171,3 +171,10 @@ func BenchmarkReplace(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkToSlicePtr(b *testing.B) {
+	preallocated := make([]int, 100000)
+	for i := 0; i < b.N; i++ {
+		_ = ToSlicePtr(preallocated)
+	}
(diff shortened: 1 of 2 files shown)


yanmhlv:
Hi @samber! Please take a look at the PR when you have time :)

samber (approved):

andig:
Out of curiosity: why does this have an effect? `Map` already pre-allocates the resulting slice- why isn't this sufficient?

PR #466, opened 2024-06-17
feat: add string conversion functions

eiixy:


string_test.go
```diff
@@ -100,3 +100,150 @@ func TestRuneLength(t *testing.T) {
...
+		args args
+		want []string
+	}{
+		{"", args{"PascalCase"}, []string{"Pascal", "Case"}},
+		{"", args{"camelCase"}, []string{"camel", "Case"}},
+		{"", args{"snake_case"}, []string{"snake", "case"}},
+		{"", args{"kebab_case"}, []string{"kebab", "case"}},
+		{"", args{"_test text_"}, []string{"test", "text"}},
+		{"", args{"test123string"}, []string{"test", "123", "string"}},
+		{"", args{"UPPERCASE"}, []string{"UPPERCASE"}},
```
ccoVeille, line 223: This test should provide real life examples, and include things about Go initialisms

Maybe you could have to look at these repositories: 

https://github.com/client9/misspell

https://github.com/ettle/strcase

string.go
```diff
@@ -94,3 +96,70 @@ func ChunkString[T ~string](str T, size int) []T {
...
+// SnakeCase converts string to snake case.
+func SnakeCase(str string) string {
+	return strings.Join(Map(Words(str), func(item string, index int) string {
+		return strings.ToLower(item)
+	}), "_")
+}
+
+// Words splits string into an array of its words.
+func Words(str string) []string {
+	reg := regexp.MustCompile(`([a-z])([A-Z])|([a-zA-Z])([0-9])|([0-9])([a-zA-Z])`)
```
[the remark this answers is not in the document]
eiixy: I agree with your point. I have optimized the regex as suggested. Please review the latest commit.

string_test.go
eiixy: I have updated the test cases

string.go
```diff
@@ -145,11 +149,17 @@ func Words(str string) []string {
...
+func Capitalize(str string) string {
+	if len(str) == 0 {
+		return str
+	}
+	runes := []rune(str)
+	for i, r := range runes {
+		if i == 0 {
+			runes[i] = unicode.ToUpper(r)
+		} else {
+			runes[i] = unicode.ToLower(r)
```
ccoVeille, line 165: This code still looks strange to me

There was a function now deprecated that did this https://pkg.go.dev/strings#Title

It had been replaced by https://pkg.go.dev/golang.org/x/text/cases#Title

string.go
```diff
@@ -145,11 +149,17 @@ func Words(str string) []string {
...
 // Capitalize converts the first character of string to upper case and the remaining to lower case.
-func Capitalize(word string) string {
-	if len(word) == 0 {
-		return word
+func Capitalize(str string) string {
+	if len(str) == 0 {
+		return str
+	}
+	runes := []rune(str)
+	for i, r := range runes {
```
ccoVeille, line 161: I would say, keep it simple

```go
for i, r := range runes {
     r = unicode.ToLower(r)
     if i == 0 {
	r = unicode.ToUpper(r)
     }
     runes[i] = r
}
```

string.go
eiixy: Thank you for your suggested changes.

samber (approved):

samber:
Thanks for the contrib. I've been looking for such functions recently \o/

NathanBaulch:
Is the comment "Dependencies are excluded from releases" in `go.mod` still true now that this feature has added a runtime dependency on "golang.org/x/text"?

string.go
```diff
@@ -1,8 +1,12 @@
 package lo
 
 import (
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"math/rand"
+	"regexp"
 	"strings"
+	"unicode"
 	"unicode/utf8"
```
ccoVeille, line 10:
```suggestion
 	"math/rand"
	"regexp"
 	"strings"
	"unicode"
 	"unicode/utf8"

	"golang.org/x/text/cases"
	"golang.org/x/text/language"
```

samber:
@NathanBaulch I corrected the comment with "dev dependencies"

PR #468, opened 2024-06-27
feat: adding Earliest and Latest

samber:


find.go
```diff
@@ -310,6 +333,28 @@ func MaxBy[T any](collection []T, comparison func(a T, b T) bool) T {
...
+func Latest(times ...time.Time) time.Time {
+	var max time.Time
+
+	if len(times) == 0 {
+		return max
+	}
+
+	max = times[0]
+
+	for i := 1; i < len(times); i++ {
```
ccoVeille, line 347: Is there any reason you use for I++ loop and not a range ?


```suggestion
for _, item := range items {
          if item.After(max) { 
                   max = item 
          }
}
```

samber: To skip the first.

Comparing a zero value with item is not that safe.

PR #470, opened 2024-06-27
feat: adding Elipse

samber:


NathanBaulch:
Curious - why the name "Elipse" vs "Ellipsis"?
This apparent typo drew my attention to this function, but while I'm here... I generally don't use this function because:
1. I prefer to append the more compact U+2026 now that Unicode is ubiquitous.
2. I like to trim trailing whitespace after truncation to ensure the ellipsis aren't floating.

ccoVeille:
I would say it's because Samuel is French (as I do) and because Elipse is the word for Ellipsis in French.

I think it's a typo

samber:
Oopsie. I'm going to fix that. 😅

samber:
I added a trim to the helper as well.

Opening an issue to discuss about the U+2026 char.

NathanBaulch:
Sorry, I should have explained "floating" ellipsis above - I meant trimming *after* truncation to avoid this scenario:

```go
is.Equal("hello...", Ellipsis("hello world", 9))
```

```
expected: "hello..."
actual  : "hello ..."
```

samber:
ok, I made another fix

PR #478, opened 2024-06-29
remove `golang.org/x/exp` since it doesn't follow go 1 compatibility promise

trim21:
[document](https://pkg.go.dev/golang.org/x/exp#section-readme) says `golang.org/x/exp` doesn't follow go 1 compatibility promise, so copy it's package we used to internal for future compatibility promise.

In the future is downsteam user import exp with a newer non-compatibility version it will get unexpected surprise.


samber:
Good to know. Thanks for the contrib!

andig:
If I were using constraints I'd be aware of them being an experimental Go feature and I'd expect libraries to follow the current /exp version. I could then also chose the /exp version I want via `replace` directives in go.mod. With this PR I can no longer do so.

samber:
The constraint defines interfaces. These interfaces are the responsibility of the receiver, not the caller.

Can you explain why you would replace the `exp` version?

andig:
> Can you explain why you would replace the exp version?

Why would you cast it in stone?

The point of fixing the `exp` version in this PR is to maintain it's current state. If I want to use a newer exp version (maybe more options for `comparable`, say complex numbers, whatever) I can no longer do so. Making this change since `lo` doesn't care what `comparable` exactly is as long as the compiler guarantees that it is this change is imho restricting flexibility.

trim21:
> The point of fixing the `exp` version in this PR is to maintain it's current state. If I want to use a newer exp version (maybe more options for `comparable`, say complex numbers, whatever) I can no longer do so. Making this change- since `lo` doesn't care what `comparable` exactly is as long as the compiler guarantees that it is- this change is imho restricting flexibility without adding any gain.

`comparable` is defined by go builtin (compiler itself), not `exp`.

`exp` define `constraints.Ordered`, it will use `cmp.Ordered` when it's available, for example, go>=1.21..

IMO the potential issue is, go may add more addable type, `constraints.Float | constraints.Integer | constraints.Complex`, for example int128. only in that case we need to update our `internal/constraints`

PR #482, opened 2024-06-30
feat: adding "mutable" subpackage

samber:


mutable/slice.go
```diff
@@ -0,0 +1,95 @@
...
+	FilterI(collection, func(item T, index int) bool {
+		return predicate(item)
+	})
+}
+
+// Filter iterates over elements of collection, returning an array of all elements predicate returns truthy for.
+// The function returns the modified slice, which may be shorter than the original if some elements were removed.
+// The order of elements in the original slice is preserved in the output.
+// Play:
+func FilterI[T any](collection *[]T, predicate func(item T, index int) bool) {
```
samber, line 19: Compared to immutable package, I chose to split `Filter` into 2 helpers: the `I` means predicate receives and index as parameter.

slice.go
```diff
@@ -273,15 +273,21 @@ func Interleave[T any, Slice ~[]T](collections ...Slice) Slice {
 // Shuffle returns an array of shuffled values. Uses the Fisher-Yates shuffle algorithm.
 // Play: https://go.dev/play/p/Qp73bnTDnc7
 func Shuffle[T any, Slice ~[]T](collection Slice) Slice {
-	rand.Shuffle(len(collection), func(i, j int) {
-		collection[i], collection[j] = collection[j], collection[i]
+	size := len(collection)
```
samber, line 276: I just figured out that `lo.Shuffle` is having a side effect.

slice.go
```diff
@@ -273,15 +273,21 @@ func Interleave[T any, Slice ~[]T](collections ...Slice) Slice {
...
 	})
 
-	return collection
+	return output
 }
 
 // Reverse reverses array so that the first element becomes the last, the second element becomes the second to last, and so on.
 // Play: https://go.dev/play/p/fhUMLvZ7vS6
+//
+// Deprecated: Use lom.Reverse instead.
```
samber, line 290: Deprecating Reverse that will be moved to samber/lo/mutable package

mutable/slice.go
```diff
@@ -0,0 +1,95 @@
+package mutable
```
samber, line 1:
```go
package mutable
```

or

```
package lom
```

?

It does matters, because it will be the default package alias on import.

mutable/slice.go
```diff
@@ -0,0 +1,95 @@
...
+		j++
+	}
+
+	*collection = (*collection)[:j]
+}
+
+// Shuffle returns an array of shuffled values. Uses the Fisher-Yates shuffle algorithm.
+// Play:
+func Shuffle[T any](collection []T) {
+	rand.Shuffle(len(collection), func(i, j int) {
```
samber, line 80: replace with `lo/internal/rand` (PR is WIP)

mutable/slice.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Filter iterates over elements of collection, returning an array of all elements predicate returns truthy for.
+// The function returns the modified slice, which may be shorter than the original if some elements were removed.
+// The order of elements in the original slice is preserved in the output.
+// Play:
+func FilterI[T any](collection *[]T, predicate func(item T, index int) bool) {
+	j := 0
+	for i := range *collection {
+		if predicate((*collection)[i], i) {
+			(*collection)[j] = (*collection)[i]
```
Elbehery, line 23: iiuc, this is `immutable`, so I think you should dump the filtered elems into a new collection, or ?

mutable/slice_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFilter(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
```
Elbehery, line 11: why not `table driven` ?

mutable/slice.go
samber: nope ;)

this subpackage would be dedicated to mutable helpers

mutable/slice_test.go
samber: what do you mean ?

Elbehery: https://go.dev/wiki/TableDrivenTests

mutable/slice.go
Elbehery: ok, it is more safe to work on a copied variable, since the iteration might mutate the collection and this could cause subtle hidden bugs iiuc 

P.S. dont vote to the right :D

samber: I'm not sure to understand. You think that `predicate()` could manipulate `collection` ?

PS: don't worry 😅✊

mutable/slice_test.go
samber: Ok 👌

I usually don't write table-driven tests for only 2 or 3 test cases.

mutable/slice.go
ccoVeille: No one will get lom, mutable as a package is good.

Don't enforce your import alias as package name 😅😬

mutable/slice.go
ccoVeille: I would not use the I name and use your current FilterI as Filter

There is an index, I don't use it, let's user ignore it. You have no need to provide a variant of all your methods

mutable/slice.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+
+	*collection = (*collection)[:j]
+}
+
+// Uniq returns a duplicate-free version of an array, in which only the first occurrence of each element is kept.
+// The order of result values is determined by the order they occur in the array.
+// Play:
+func Uniq[T comparable](collection *[]T) {
```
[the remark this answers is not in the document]
trim21: yes there is already a stdlib version of this， don't add it.

mutable/slice.go
```diff
@@ -0,0 +1,95 @@
...
+		(*collection)[j] = (*collection)[i]
+		j++
+	}
+
+	*collection = (*collection)[:j]
+}
+
+// Shuffle returns an array of shuffled values. Uses the Fisher-Yates shuffle algorithm.
+// Play:
+func Shuffle[T any](collection []T) {
```
trim21, line 79: I would suggest to not add this methos, it's just save 2 lines saver alias for stdlib.

samber: The point of adding this function here is to deprecate it from parent package.

2y ago, I added 2 functions with mutable behavior (Shuffle + Reverse). It was an error.

mutable/slice.go
Elbehery:
```
(*collection)[j] = (*collection)[i]
```

mutable/slice_test.go
Elbehery: i used to do the same, but i always got reviews about it.. so if it is one case only then ok, otherwise always use tables

mutable/slice.go
ccoVeille: Add a `// Deprecated ` note then, if you don't want them anymore

PR #485, opened 2024-07-02
add ForEachCondition implement

Sianao:
In certain cases, we want to control whether the foreach continues to execute, such as checking if a set of data meets certain criteria. ForEachCondition implements this functionality to address this need

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -94,6 +94,16 @@ func ForEach[T any](collection []T, iteratee func(item T, index int)) {
 	}
 }
 
+// ForEachCondition iterates over elements of collection and invokes iteratee for each element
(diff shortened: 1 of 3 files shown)


samber:
Good one! ✌️

I wonder if we should rename it to ForEachWhile? 🤔 It would harmonize the name of helpers that finish executing with a condition: DropWhile, AttemptWhile.

Feel free to argue about it, or propose another name!

samber (approved):

Sianao:
Sir , is any problem with this pull request ?

printercu:
I see a lot of new features landing recently. What is the benefit of using this function over "for i := range"?

PR #489, opened 2024-07-12
feat: adding EarliestBy and LatestBy functions

timych:
This query adds two functions EarliestBy and LatestBy, which are helpful when dealing with slice of structures or more complex time calculation cases.


samber (approved):

samber:
LGTM, thanks for this contrib ;)

Issue #438, opened 2024-04-02
Chunk has memory leak


shoham-b:
Hi.
Appending to a one of the chunk resulted from `lo.Chunk` might leak into to consecutive chunk.
```go
func TestChunk(t *testing.T) {
	allData := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
	chunkedData := lo.Chunk(allData, 5)
	firstChunk := chunkedData[0]
	secondChunk := chunkedData[1]
	firstChunk = append(firstChunk, 666)
	fmt.Printf("secondChunk: %v\n", secondChunk) // secondChunk: [666 7 8 9 10]
}
```

The reason for that is that the slicing does not add the capacity, so the default max capacity is used, and then upon append the slice isn't reallocated.

To fix that we simply have to slice by 3 parts
```go
		result = append(result, collection[i*size:last:last])
```
rather then the current implementation of
```go
		result = append(result, collection[i*size:last]) // Bug
```

PR #491, opened 2024-07-15
fix: chunk memory leak, bug fix

mihir20:
raising a fix as mentioned in: https://github.com/samber/lo/issues/438

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -182,7 +182,7 @@ func Chunk[T any, Slice ~[]T](collection Slice, size int) []Slice {
 		if last > len(collection) {
 			last = len(collection)
 		}
-		result = append(result, collection[i*size:last])
+		result = append(result, collection[i*size:last:last])
 	}
 
 	return result
--- a/slice_test.go
+++ b/slice_test.go
@@ -235,6 +235,12 @@ func TestChunk(t *testing.T) {
 	allStrings := myStrings{"", "foo", "bar"}
 	nonempty := Chunk(allStrings, 2)
 	is.IsType(nonempty[0], allStrings, "type preserved")
+
+	// appending to a chunk should not affect original array
+	originalArray := []int{0, 1, 2, 3, 4, 5}
+	result5 := Chunk(originalArray, 2)
+	result5[0] = append(result5[0], 6)
+	is.Equal(originalArray, []int{0, 1, 2, 3, 4, 5})
 }
 
 func TestPartitionBy(t *testing.T) {


slice.go
```diff
@@ -182,7 +182,7 @@ func Chunk[T any, Slice ~[]T](collection Slice, size int) []Slice {
 		if last > len(collection) {
 			last = len(collection)
 		}
-		result = append(result, collection[i*size:last])
+		result = append(result, collection[i*size:last:last])
```
ccoVeille, line 185: I had never faced this notation, with the capacity apparently

slice_test.go
```diff
@@ -235,6 +235,12 @@ func TestChunk(t *testing.T) {
 	allStrings := myStrings{"", "foo", "bar"}
 	nonempty := Chunk(allStrings, 2)
 	is.IsType(nonempty[0], allStrings, "type preserved")
+
+	// appending to a chunk should not affect original array
+	originalArray := []int{0, 1, 2, 3, 4, 5}
+	result5 := Chunk(originalArray, 2)
+	result5[0] = append(result5[0], 6)
+	is.Equal(originalArray, []int{0, 1, 2, 3, 4, 5})
```
ccoVeille, line 243: So. it's not only a fix of a possible memory leak, but also a functional fix, no?

I'm unsure why you are adding this check.no computer with me to check.

My point is that the commit message could mention a memory leak fix + bug fix, maybe

ccoVeille (approved):
LGTM 👍

slice_test.go
mihir20: yeah, updated the title

slice.go
samber, line 185: Why not `collection[i*size:last:size]` ?

slice_test.go
ccoVeille: Thanks

samber:
I would recommend using `copy()` instead of a pointer to the same memory space. I think this is the spirit of this library. WDYT?

It consumes more memory, but it seems safer to me.

mihir20:
> I would recommend using `copy()` instead of a pointer to the same memory space. I think this is the spirit of this library. WDYT?
> 
> It consumes more memory, but it seems safer to me.

using copy will be slower compared to current implementation:
Benchmark using copy:
```
BenchmarkChunk
BenchmarkChunk/strings_10
BenchmarkChunk/strings_10-12         	14886344	        80.03 ns/op
BenchmarkChunk/strings_100
BenchmarkChunk/strings_100-12        	 1814036	       660.8 ns/op
BenchmarkChunk/strings_1000
BenchmarkChunk/strings_1000-12       	  183382	      6143 ns/op
BenchmarkChunk/ints10
BenchmarkChunk/ints10-12             	20346310	        57.41 ns/op
BenchmarkChunk/ints100
BenchmarkChunk/ints100-12            	 2672658	       433.6 ns/op
BenchmarkChunk/ints1000
BenchmarkChunk/ints1000-12           	  312517	      3871 ns/op
```
Benchmark without using copy:
```
BenchmarkChunk
BenchmarkChunk/strings_10
BenchmarkChunk/strings_10-12         	52708929	        22.42 ns/op
BenchmarkChunk/strings_100
BenchmarkChunk/strings_100-12        	12681414	        93.45 ns/op
BenchmarkChunk/strings_1000
BenchmarkChunk/strings_1000-12       	 2022960	       599.6 ns/op
BenchmarkChunk/ints10
BenchmarkChunk/ints10-12             	52086441	        22.50 ns/op
BenchmarkChunk/ints100
BenchmarkChunk/ints100-12            	12925526	        90.06 ns/op
BenchmarkChunk/ints1000
BenchmarkChunk/ints1000-12           	 1978053	       591.1 ns/op
```

slice.go
mihir20: ref: https://stackoverflow.com/questions/12768744/re-slicing-slices-in-golang/18911267#18911267

samber:
I will discuss it in another issue.

Issue #467, opened 2024-06-18
proposal: Another PickByKeys that ouputs a slice


MakDon:
# Example

```
m := lo. PickByKeys(map[string]int{"foo": 1, "bar": 2, "baz": 3}, []string{"foo", "baz"})
// []int{1,3}
```

The output values' order should be the same as the input slice. 

# Why we need it

Assume we have a list of openids:
```
openids = []string{openid1, openid2}
```

Then we have a method to transfer it into user_id:
```
func ToUserIDs(openids []string) map[string]string{}
```
The output of `ToUserIDs` is a map, in which the key is openid and the value is UserID.

If I call `Values`, then i would get a unordered list, which can not keep the original order with `openids`

So maybe we need one "Ordered PickByKeys function that returns a slice"?

Further more, the method could support the following scene:

```golang
	gourpA := []string{openid1, openid2, openid3}
	gourpB := []string{openid4, openid5, openid6}
	gourpC := []string{openid7, openid8, openid9}
	all := append([]string{}, gourpA...)
	all := append(all, gourpB...)
	all := append(all, gourpC...)
	users := fetchUsers(all)
	usersA := lo.PickByKeys(users, gourpA)
	usersB := lo.PickByKeys(users, gourpB)
	usersC := lo.PickByKeys(users, gourpC)
```

Join all the IDs, fetch in one request, then separate the returned entities into different groups.

PR #494, opened 2024-07-16
feat<map> add ValuesByKeys for get values array in order [issue - 467]

XiaKuan:
solve 
#467

### DIFF
--- a/map.go
+++ b/map.go
@@ -76,6 +76,17 @@ func PickByValues[K comparable, V comparable, Map ~map[K]V](in Map, values []V) 
 	return r
 }
 
+// ValuesByKeys returns an array of values in the same order as the given keys.
+func ValuesByKeys[K comparable, V any](in map[K]V, keys []K) []V {
+	out := make([]V, 0, len(keys))
+	for i := range keys {
+		if v, ok := in[keys[i]]; ok {
+			out = append(out, v)
+		}
+	}
+	return out
+}
+
 // OmitBy returns same map type filtered by given predicate.
 // Play: https://go.dev/play/p/EtBsR43bdsd
 func OmitBy[K comparable, V any, Map ~map[K]V](in Map, predicate func(key K, value V) bool) Map {
--- a/map_example_test.go
+++ b/map_example_test.go
@@ -66,6 +66,15 @@ func ExamplePickByValues() {
 	// Output: 2 1 3
 }
 
+func ExampleValuesByKeys() {
+	kv := map[string]int{"foo": 1, "bar": 2, "baz": 3}
+
+	result := ValuesByKeys(kv, []string{"baz", "foo", "bar"})
+
+	fmt.Printf("%v %v %v", len(result), result[0], result[1])
+	// Output: 3 3 1
+}
+
 func ExampleOmitBy() {
 	kv := map[string]int{"foo": 1, "bar": 2, "baz": 3}
 
--- a/map_test.go
+++ b/map_test.go
@@ -95,6 +95,19 @@ func TestPickByValues(t *testing.T) {
 	is.IsType(after, before, "type preserved")
 }
 
+func TestValuesByKeys(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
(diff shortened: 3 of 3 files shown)


XiaKuan:
This issue can also be resolved in this way
```go
func main() {
    inputMap := map[string]int{"foo": 1, "bar": 2, "baz": 3}
    inputSlice := []string{"foo", "baz"}
    result := lo.Map(inputSlice, func(key string, _ int) int {
        return inputMap[key]
    })
}
```

XiaKuan:
This issue can also be resolved in this way
```go
func main() {
    inputMap := map[string]int{"foo": 1, "bar": 2, "baz": 3}
    inputSlice := []string{"foo", "baz"}
    result := lo.Map(inputSlice, func(key string, _ int) int {
        return inputMap[key]
    })
}
```
But I think this is used more.
Is it recommended to add this API?

map.go
```diff
@@ -76,6 +76,18 @@ func PickByValues[K comparable, V comparable, Map ~map[K]V](in Map, values []V)
 	return r
 }
 
+// ValuesByKeys returns an array of values in the same order as the given keys.
+// Play: https://go.dev/play/p/PmmPDo1AqWl
+func ValuesByKeys[K comparable, V any](in map[K]V, keys []K) []V {
+	out := make([]V, 0, len(keys))
+	for i := range keys {
+		if v, ok := in[keys[i]]; ok {
```
shivamrazorpay, line 84: Just a random doubt: If the key is not present in the map, then the order in the output slice will not be aligned to the input keys.
You can add that edge case in the example of Readme. 
Your thoughts on it ??

shivamrazorpay:
Hi, can you review it @samber ?

map.go
XiaKuan: Thanks for your suggestion!
I agree with you to build this edge case into the example. 
Another way would be to append nil or zero values to the slice, but this solution requires additional checking by the caller, which I don't think is a good practice.
WDYT?

ccoVeille: I agree with @shivamrazorpay 

It's the first thing I thought about

I would expect to return an error in such case

README.md
```diff
@@ -1120,6 +1121,18 @@ m := lo.PickByValues(map[string]int{"foo": 1, "bar": 2, "baz": 3}, []int{1, 3})
 
 [[play](https://go.dev/play/p/1zdzSvbfsJc)]
 
+### ValuesByKeys
+
+Returns an array of values in the same order as the given keys.
+If the key in keys slice but not in the map, return error
+```go
+
+m := lo.ValuesByKeys(map[string]int{"foo": 1, "bar": 2, "baz": 3}, []string{"baz", "foo", "bar"})
```
ccoVeille, line 1130:
```suggestion
m, err := lo.ValuesByKeys(map[string]int{"foo": 1, "bar": 2, "baz": 3}, []string{"baz", "foo", "bar"})
```

map.go
XiaKuan: That seems reasonable.

I have resubmitted the code

map.go
```diff
@@ -76,6 +80,21 @@ func PickByValues[K comparable, V comparable, Map ~map[K]V](in Map, values []V)
...
+// Play: https://go.dev/play/p/PmmPDo1AqWl
+func ValuesByKeys[K comparable, V any](in map[K]V, keys []K) ([]V, error) {
+	out := make([]V, 0, len(keys))
+	for i := range keys {
+		if v, ok := in[keys[i]]; ok {
+			out = append(out, v)
+		} else {
+			return nil, fmt.Errorf("ValuesByKeys: %v is not in the map", keys[i])
+		}
+	}
```
ccoVeille, line 94:
```suggestion
	for i := range keys {
		v, ok := in[keys[i]]
		if !ok {
			return nil, fmt.Errorf("ValuesByKeys: %v is not in the map", keys[i])
		}
		out = append(out, v)
	}
```

ccoVeille (approved):
LGTM 👍

samber:
Instead of returning error, I would return the list of missing keys.

Also, i'm not even sure that returning a second argument is suitable. Because such helpers cannot be chained or used in argument to a function.

PR #495, opened 2024-07-16
Add EmptyOnError, EmptyOnError{1->6}

yuweiweiouo:
## Summary
This PR introduces a new helper function named `EmptyOnError` along with `EmptyOnError1` to `EmptyOnError6`. These functions are designed to safely retrieve values while ignoring errors, ensuring that in the case of an error, empty values are returned instead of triggering a panic.

## Motivation
In many scenarios, developers may only be interested in retrieving a value and are willing to ignore potential errors without risking a panic. The existing `Must` helper function, while useful, panics when an error is encountered, which might not always be desirable. The `EmptyOnError` helper and its variants (`EmptyOnError1` to `EmptyOnError6`) provide a safer alternative by returning empty values in case of an error.

Unlike `Must`, which can handle zero return values, `EmptyOnError` focuses solely on the values returned by a function. Therefore, there is no `EmptyOnError0` variant, as it would be redundant.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -113,7 +113,7 @@ func Must6[T1, T2, T3, T4, T5, T6 any](val1 T1, val2 T2, val3 T3, val4 T4, val5 
 	return val1, val2, val3, val4, val5, val6
 }
 
-// just assume fine if err is nil or true.
+// fine assume ok if err is nil or true.
 func fine(err any) bool {
 	if err == nil {
 		return true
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -113,6 +113,76 @@ func Must6[T1, T2, T3, T4, T5, T6 any](val1 T1, val2 T2, val3 T3, val4 T4, val5
 	return val1, val2, val3, val4, val5, val6
 }
 
+// just assume fine if err is nil or true.
+func fine(err any) bool {
```
ccoVeille, line 117: I won't use `err` as a variable name here, as the type is unknown.

I would update the signature of all Just* methods

But it's the way must was written, so …

samber:
Hi @yuweiweiouo and thanks for your first contribution.

About the naming, At first, I thought that "Just" was referring to the reactive programming helper that creates a stream of N elements.

Can you suggest other names?

- EmptyOnError
- other?

errors.go
yuweiweiouo: This is a reasonable suggestion,  maybe I can adjust `must` along with it.

yuweiweiouo: How do you feel about changing `err` to `failure`?

ccoVeille: Naming is a pain.

@samber is better than I am 😬😂

yuweiweiouo:
> Hi @yuweiweiouo and thanks for your first contribution.
> 
> About the naming, At first, I thought that "Just" was referring to the reactive programming helper that creates a stream of N elements.
> 
> Can you suggest other names?
> 
> * EmptyOnError
> * other?

I really like short and strong names like `Must`. How about changing the name to `Silect` or `Suppress`?

yuweiweiouo:
Here are a few candidates that I prefer:
* Fine ( just fine )
* May ( related to Must )
* Silence
* Suppress

yuweiweiouo:
Hi @samber 
I've gone with `Fine`
WDYT?

errors.go
samber: I have a good friend named ChatGPT. Perfect for brainstorming.

yuweiweiouo:
Hi @samber,
I've changed it to the `EmptyOnError`, Could you kindly review it?

ccoVeille:
Was it merged or abandoned?

yuweiweiouo:
> Was it merged or abandoned?

Since it has been idle for too long and there has been no response, I believe this PR will not be accepted, so I closed it for now

PR #496, opened 2024-07-17
Update string.go

liujundezhanghao:
more reasonable

### DIFF
--- a/string.go
+++ b/string.go
@@ -58,7 +58,7 @@ func Substring[T ~string](str T, offset int, length uint) T {
 		}
 	}
 
-	if offset > size {
+	if offset >= size {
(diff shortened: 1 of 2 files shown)


ccoVeille:
Could you please add test for this use case?

It's unclear what you are fixing, and someone could open a PR and make the opposite changes

liujundezhanghao:
Because if the offsets are equal in size, the following logic is meaningless.

liujundezhanghao:
Thanks, I added it

ccoVeille:
Thanks for doing the changes. Let's wait for maintainers feedback now

samber:
👌 Good point

PR #498, opened 2024-07-18
Feat: Adding the equivalent of checking if two slices are the same

Dokiys:
Often, I need to compare if two slices are same in elements and count of each element. However, I have not found an existing method that accomplishes this functionality. Therefore, I have added this method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Supported intersection helpers:
 - [ContainsBy](#containsby)
 - [Every](#every)
 - [EveryBy](#everyby)
-- [Equivalent](#equivalent)
+- [EqualUnordered](#equalunordered)
 - [Some](#some)
 - [SomeBy](#someby)
 - [None](#none)
@@ -1951,7 +1951,7 @@ b := EveryBy([]int{1, 2, 3, 4}, func(x int) bool {
 ### EqualUnordered
 
 Returns true if the subset has the same elements and the same number of each element as the collection.
-Unlike slices.Equal(), which returns effected by order, Equivalent does not care about the order of the elements.
+Unlike slices.Equal(), which returns effected by order, EqualUnordered does not care about the order of the elements.
 
 ```go
 b := EqualUnordered([]int{0, 1, 1, 3, 0, 0}, []int{0, 1, 3, 0, 1, 0})
--- a/README.md
+++ b/README.md
@@ -1948,13 +1948,13 @@ b := EveryBy([]int{1, 2, 3, 4}, func(x int) bool {
 // true
 ```
 
-### Equivalent
+### EqualUnordered
 
 Returns true if the subset has the same elements and the same number of each element as the collection.
 Unlike slices.Equal(), which returns effected by order, Equivalent does not care about the order of the elements.
 
 ```go
-b := Equivalent([]int{0, 1, 1, 3, 0, 0}, []int{0, 1, 3, 0, 1, 0})
+b := EqualUnordered([]int{0, 1, 1, 3, 0, 0}, []int{0, 1, 3, 0, 1, 0})
 // true
 ```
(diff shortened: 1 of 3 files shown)


ccoVeille:
Your tests are focusing on slices

Did you take a look at this?

https://pkg.go.dev/slices#Equal

intersect.go
```diff
@@ -44,6 +44,28 @@ func EveryBy[T any](collection []T, predicate func(item T) bool) bool {
...
+		m[collection[i]] += 1
+	}
+	for i := range subset {
+		m[subset[i]] -= 1
+	}
+	for _, v := range m {
+		if v != 0 {
+			return false
+		}
+	}
```
shivamrazorpay, line 65: Your logic requires extra space in the map and also 3 times iterating over the map.
Why not just sort the 2 arrays and compare if both are the same ?? 

What are your thoughts ??

Also after adding the function, you can add the function to Readme also.
Look this PR as reference : https://github.com/samber/lo/pull/494

Dokiys:
> Your tests are focusing on slices
> 
> Did you take a look at this?
> 
> https://pkg.go.dev/slices#Equal

However, slice.Equal() return is effected by elements order.  In my case, order is dosen't matter.

ccoVeille:
> > Your tests are focusing on slices
> > 
> > Did you take a look at this?
> > 
> > https://pkg.go.dev/slices#Equal
> 
> However, slice.Equal() return is effected by elements order.  In my case, order is dosen't matter.

Indeed, you may add this in the comment of method, especially how it differs from slice.Equal

intersect.go
Dokiys: Because just `!=` or `==` is allowed on generics comparable, we can't sort this slice. 
Additionally, sort.Slice() will change the parameter's original order, which I think is unsafe.

Dokiys:
> > > Your tests are focusing on slices
> > > Did you take a look at this?
> > > https://pkg.go.dev/slices#Equal
> > 
> > 
> > However, slice.Equal() return is effected by elements order.  In my case, order is dosen't matter.
> 
> Indeed, you may add this in the comment of method, especially how it differs from slice.Equal

I have added the comments based on your suggestion. Additionally, I have modified the generic declaration to mirror slice.Equal. :)

samber:
Thanks @Dokiys for your first contribution

I agree with @ccoVeille: this helper's name is not self-explaining. Some ideas: EqualUnorderedSlice, EqualUnordered ("slice" is implicit because maps cannot be ordered)...

Also, can the last argument be converted into a vaarg? It might slow it down at some point. Feel free to argue against it if your benchmarks show it is a bad idea.

README.md
```diff
@@ -196,6 +196,7 @@ Supported intersection helpers:
 - [ContainsBy](#containsby)
 - [Every](#every)
 - [EveryBy](#everyby)
+- [Equivalent](#equivalent)
```
ccoVeille, line 199: You have to fix this

Dokiys:
> Thanks @Dokiys for your first contribution
> 
> I agree with @ccoVeille: this helper's name is not self-explaining. Some ideas: EqualUnorderedSlice, EqualUnordered ("slice" is implicit because maps cannot be ordered)...
> 
> Also, can the last argument be converted into a vaarg? It might slow it down at some point. Feel free to argue against it if your benchmarks show it is a bad idea.

Thanks for your suggestions. I agree that using the name EqualUnordered would indeed be more appropriate. 

However, regarding the use of vaarg, I have reservations. This isn't based on performance issues.
I envision this method being used to compare two slices specifically, rather than comparing 'a certain slices' with 'certain data'. 
Additionally, I've referenced some methods in the `slices` package and noticed that variadic arguments are almost never used there either.

PR #499, opened 2024-07-18
Feature/partition in 2

dsolerh:
I find PartitionIn2By useful in the scenarios that I described in the doc comment of the function.
also this is my first PR into this repo, let me know if anything is not complaint with the style, I'm more than willing to fix it.


dsolerh:
I totally agree with you, I didn't saw the FilterReject before, that's on me, thanks for the time dedicated to this.

PR #501, opened 2024-07-18
ADDED: Nillable function 

Nizom98:
This function helps you not to check the input parameter for nil every time.

### DIFF
--- a/type_manipulation.go
+++ b/type_manipulation.go
@@ -120,3 +120,11 @@ func CoalesceOrEmpty[T comparable](v ...T) T {
 	result, _ := Coalesce(v...)
 	return result
 }
+
+// Nillable return [nil] if [v] is nil, otherwise returns [fn] call result.
+func Nillable[T, RT any](v *T, fn func(*T) *RT) *RT {
+	if IsNil(v) {
+		return Nil[RT]()
+	}
+	return fn(v)
+}
--- a/type_manipulation_test.go
+++ b/type_manipulation_test.go
@@ -294,3 +294,37 @@ func TestCoalesceOrEmpty(t *testing.T) {
 	is.Equal(result9, struct1)
 	is.Equal(result10, struct1)
 }
+
+func TestNillable(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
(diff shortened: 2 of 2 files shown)


samber:
Why are you limiting this helper to pointers?

The current `lo.Coalesce` helper returns the first non-zero argument.

I wonder if you should call it `lo.CoalesceF` instead. What about chaining a list of callbacks with vaargs?

If your first argument is a value instead of func, it would be a good use case for the `lo.Identity` helper suggested in #420.

```go
// values
lo.Coalesce(0, 42, 21) // 42

// func
lo.CoalesceF(returnsZero, returnsFortyTwo, returnsTwentyOne) // 42

// mix
lo.CoalesceF(returnsZero, lo.Identity(42), returnsTwentyOne) // 42
```

WDYT ?

Nizom98:
Than you for your comment. 
> The current lo.Coalesce helper returns the first non-zero argument.

The goal of **Nillable** function is not returning non-nill argument. The goal is simplify converting between structs. 
For example 

```
type type1 struct {
	field1 string
}
type type2 struct {
	field2 string
}


// I need convert **type1** fields into **type2**.
// The first way I can do is write "convert" function for all types.
type2Val := convType1ToType2(type1Val)
func convType1ToType2(in *type1) *type2 {
	if IsNil(nil) {
		return nil
	}
	return &type2{field2: in.field1}
}


// The second way to do this is the Nillable function.
// In this case I do not need check every time, that the inout parameter is **_nil_** and I do not write converter function for all types. 
type2Val := Nillable(type1Val, func(in *type1) *type2 {
	return &type2{field2: in.field1}
})

```

Nizom98:
> Why are you limiting this helper to pointers?
I think, you are right here. We do no need limit this helper to pointers. I will fix it.

Nizom98:
> Why are you limiting this helper to pointers?

If we will use non-pointer type, I could not determine default value to return. 
Here is an example. 
```go
func Nillable[T, RT any](v T, fn func(T) RT) RT {
	if IsNil(v) {
		// TODO: should we return new(RT) or *new(RT)?
	}
	return fn(v)
}
```

samber:
I think what you need is https://github.com/samber/mo/

Issue #452, opened 2024-05-13
Keys: accept multiple maps (vaargs)


samber:
From

```go
lo.Keys[K comparable, V any](in map[K]V) []K
```

to

```go
lo.Keys[K comparable, V any](in ...map[K]V) []K
```

In case of collision, deduplicate keys?

shivamrazorpay:
Hey @samber I want to add a PR to add. Can I get access to adding a PR for this issue ??

Also, I can see the Test of Keys is incorrect as it sorts the keys before asserting them, but there is no sort function in the function. I will also fix them.

shivamrazorpay:
@ccoVeille I want to be a contributor. Can you help me to get access? I can fix this issue and some other open issues.
Thank you. 😊

ccoVeille:
I don't have much access than you do I think. I'm just a compulsive code reviewer. That's why you see me active.

You just have to fork the repository, push a branch on it, then open a PR mentioning the issue you want to fix.

I can review a PR if you open one.

shivamrazorpay:
@ccoVeille by seeing your code reviews, I was amazed at how much good you are doing. I would love to chat with you one day. Can you also review my PR?
Thank you.

ccoVeille:
you are nice, thanks.

You can join the slack workspace of the gophers. You will find the link on my GitHub profile

dagenius007:
Hello @ccoVeille how do I join the gophers slack channel?

ccoVeille:
Click on the link I provided, create an account if needed, install slack if needed. 

If it doesn't work, try with this link

https://join.slack.com/t/gophers/shared_invite/zt-2lmem2ic2-EfbtKK9FGlBHHgeEu0an7Q

Luigi31415:
@samber This was resolved in #503, I think we should close this issue

PR #503, opened 2024-07-19
Keys+Values: accept multiple maps (vaargs) - Adding UniqKeys+UniqValues

shivamrazorpay:
Resolved Issue : https://github.com/samber/lo/issues/452

Added : 

- Accept Multiple maps as vaargs in Keys
- Accept Multiple maps as vaargs in Values
- Added UniqKeys which creates an array of unique map keys. 
- Added UniqValues which creates an array of unique map values. 
- Added Test for the same

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,11 +125,11 @@ Supported helpers for slices:
 Supported helpers for maps:
 
 - [Keys](#keys)
-- [UniqKeys](#UniqKeys)
-- [HasKey](#HasKey)
+- [UniqKeys](#uniqKeys)
+- [HasKey](#hasKey)
 - [ValueOr](#valueor)
 - [Values](#values)
-- [UniqValues](#UniqValues)
+- [UniqValues](#uniqValues)
 - [PickBy](#pickby)
 - [PickByKeys](#pickbykeys)
 - [PickByValues](#pickbyvalues)
--- a/README.md
+++ b/README.md
@@ -125,11 +125,11 @@ Supported helpers for slices:
 Supported helpers for maps:
 
 - [Keys](#keys)
-- [UniqKeys](#keys)
+- [UniqKeys](#uniqkeys)
 - [HasKey](#HasKey)
 - [ValueOr](#valueor)
 - [Values](#values)
-- [UniqValues](#keys)
+- [UniqValues](#uniqvalues)
 - [PickBy](#pickby)
 - [PickByKeys](#pickbykeys)
 - [PickByValues](#pickbyvalues)
--- a/README.md
+++ b/README.md
@@ -1090,8 +1090,8 @@ exists := lo.HasKey(map[string]int{"foo": 1, "bar": 2}, "baz")
 ### Values
 
 Creates an array of the map values.
-> [!NOTE]
-> The order of the values isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
+
+Use the UniqValues variant to deduplicate common values.
 
 ```go
 values := lo.Values(map[string]int{"foo": 1, "bar": 2})
@@ -1109,8 +1109,6 @@ values := lo.Values(map[string]int{"foo": 1, "bar": 2}, map[string]int{"bar": 2}
 ### UniqValues
 
 Creates an array of unique map values.
-
-Use the UniqValues variant to deduplicate common values.
 
 ```go
 values := lo.Values(map[string]int{"foo": 1, "bar": 2})
--- a/README.md
+++ b/README.md
@@ -1043,6 +1043,9 @@ Creates an array of the map keys.
 ```go
 keys := lo.Keys(map[string]int{"foo": 1, "bar": 2})
 // []string{"foo", "bar"}
+
+keys := lo.Keys(map[string]int{"foo": 1, "bar": 2}, map[string]int{"baz": 3})
+// []string{"foo", "bar", "baz"}
 ```
 
 [[play](https://go.dev/play/p/Uu11fHASqrU)]
--- a/README.md
+++ b/README.md
@@ -1041,8 +1041,8 @@ result = lo.Splice([]string{"a", "b"}, 42, "1", "2")
 ### Keys
 
 Creates a slice of the map keys.
-> [!NOTE]
-> The order of the keys isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
+
+Use the UniqKeys variant to deduplicate common keys.
 
 ```go
 keys := lo.Keys(map[string]int{"foo": 1, "bar": 2})
@@ -1060,8 +1060,6 @@ keys := lo.Keys(map[string]int{"foo": 1, "bar": 2}, map[string]int{"bar": 3})
 ### UniqKeys
 
 Creates an array of the unique map keys. 
-> [!NOTE]
-> The order of the keys isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
 
 ```go
 keys := lo.Keys(map[string]int{"foo": 1, "bar": 2}, map[string]int{"baz": 3})
@@ -1108,9 +1106,9 @@ values := lo.Values(map[string]int{"foo": 1, "bar": 2}, map[string]int{"bar": 2}
 
 ### UniqValues
 
-Creates an array of the unique map values.
-> [!NOTE]
-> The order of the values isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
+Creates an array of unique map values.
+
+Use the UniqValues variant to deduplicate common values.
 
 ```go
 values := lo.Values(map[string]int{"foo": 1, "bar": 2})
--- a/README.md
+++ b/README.md
@@ -1042,7 +1042,7 @@ result = lo.Splice([]string{"a", "b"}, 42, "1", "2")
 
 Creates a slice of the map keys.
 > [!NOTE]
-> The order of the keys isn't guaranteed to be sorted, so sort the output slice if needed.
+> The order of the keys isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
 
 ```go
 keys := lo.Keys(map[string]int{"foo": 1, "bar": 2})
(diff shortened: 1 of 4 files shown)


ccoVeille (changes requested):
Coding is not only about coding the things you need to add, but to think about all the use cases your additional features could lead to.

map_test.go
```diff
@@ -14,9 +14,13 @@ func TestKeys(t *testing.T) {
 	is := assert.New(t)
 
 	r1 := Keys(map[string]int{"foo": 1, "bar": 2})
-	sort.Strings(r1)
```
ccoVeille, line 17: You must keep the sort, or add it to Keys function, because a map is unordered while a slice is. So the way the loop is made on the map is random, so the order in the slide is random

shivamrazorpay: Got it. My bad this PR made me learn a lot. That is why we had sort in the first place in the test functions.
Sure I have added it back.
And as for sorting in the Keys() function, let's not add it. It will just add more time complexity to the function. If any user wants to get it sorted. They can sort it after the slice is returned (depending on there usecase)

What are your thoughts ??

map.go
```diff
@@ -2,13 +2,13 @@ package lo
...
-func Keys[K comparable, V any](in map[K]V) []K {
-	result := make([]K, 0, len(in))
-
-	for k := range in {
-		result = append(result, k)
+func Keys[K comparable, V any](in ...map[K]V) []K {
+	result := make([]K, 0)
+	for i := range in {
+		for k := range in[i] {
+			result = append(result, k)
```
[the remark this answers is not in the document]
shivamrazorpay: Sure , i have modified to return only unique values & added test for the same.

map.go
```diff
@@ -2,13 +2,13 @@ package lo
 
 // Keys creates an array of the map keys.
 // Play: https://go.dev/play/p/Uu11fHASqrU
-func Keys[K comparable, V any](in map[K]V) []K {
-	result := make([]K, 0, len(in))
-
-	for k := range in {
-		result = append(result, k)
+func Keys[K comparable, V any](in ...map[K]V) []K {
```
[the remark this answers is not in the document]
shivamrazorpay: Added test for his case. Thank you for the suggestion of this test.

ccoVeille (approved):
LGTM 👍 thanks

Let's wait for @samber feedback now

samber:
According to your changes, you're doing 2 operations: list keys and dedup.

Can you argue on this behavior?

IMO, some developers would only need to list keys, eg: `lo.CountValues(lo.Keys(a, b))`

Others might want to list keys with deduplication, but it could be done with 2 methods:
- `lo.Uniq(lo.Keys(a, b))`
- or a faster implem: `lo.UniqKeys(a, b)`

Also, I'm thinking about a very similar helper: `lo.Values`. Would you dedup values? I don't think so.

So to push this work forward, I would suggest the following:
- vaarg on lo.Keys without dedup
- a new helper lo.UniqKeys
- vaarg on lo.Values without dedup
- a new helper lo.UniqValues

Any ideas on the matter?

shivamrazorpay:
@samber seems like a fair callout (I was also thinking along the same lines) and also a good suggestion. I will make the 4 changes you have asked and will ask you to review them. Thank you for your suggestion.

samber:
Sure. I'm also curious about other developer's opinions on this.

ccoVeille:
I'm fine with the changes @samber suggest

I just want to make sure the code will be benchmarked against slices.Compact when using in-house dedup once you will get the keys or values

shivamrazorpay:
@samber @ccoVeille added the required changes. 
can you please review it? 🙏

map.go
```diff
@@ -2,11 +2,31 @@ package lo
...
+func UniqKeys[K comparable, V any](in ...map[K]V) []K {
+	seen := make(map[K]bool)
+	result := make([]K, 0)
+
+	for i := range in {
+		for k := range in[i] {
+			if _, exists := seen[k]; !exists {
+				seen[k] = true
+				result = append(result, k)
+			}
```
ccoVeille, line 28:
```suggestion
			if _, exists := seen[k]; exists {
				continue
			}
			seen[k] = true
			result = append(result, k)
```

map.go
```diff
@@ -21,11 +41,31 @@ func HasKey[K comparable, V any](in map[K]V, key K) bool {
...
+	result := make([]V, 0)
+
+	for i := range in {
+		for k := range in[i] {
+			if _, exists := seen[in[i][k]]; !exists {
+				seen[in[i][k]] = true
+				result = append(result, in[i][k])
+			}
+		}
 	}
```
ccoVeille, line 69: Using a temporary variable may improve the readability

```suggestion
	for i := range in {
		for k := range in[i] {{
			key := in[i][k]
			if _, exists := seen[key]; !exists {
				continue
			}
			seen[key] = true
			result = append(result, key)
		}
 	}
```

Please note the remark on map[K]struct{} may apply

map_example_test.go
```diff
@@ -9,18 +9,42 @@ import (
 
 func ExampleKeys() {
 	kv := map[string]int{"foo": 1, "bar": 2}
+	kv2 := map[string]int{"baz": 3}
 
-	result := Keys(kv)
+	result := Keys(kv, kv2)
+	sort.StringSlice(result).Sort()
```
ccoVeille, line 15: Why not using sort.String as you do in tests?

README.md
```diff
@@ -1038,15 +1040,39 @@ result = lo.Splice([]string{"a", "b"}, 42, "1", "2")
...
+// []string{"foo", "bar", "bar"}
 ```
 
 [[play](https://go.dev/play/p/Uu11fHASqrU)]
 
+### UniqKeys
+
+Creates an array of the unique map keys. 
+> [!NOTE]
+> The order of the keys isn't guaranteed to be sorted, so sort the output slice if needed.
```
ccoVeille, line 1064:
```suggestion
> The order of the keys isn't guaranteed to be sorted for performance reason, so sort the output slice if needed.
```

map.go
```diff
@@ -2,11 +2,31 @@ package lo
...
+// Play: https://go.dev/play/p/RyIE6Pb5dVS
+func UniqKeys[K comparable, V any](in ...map[K]V) []K {
+	seen := make(map[K]struct{})
+	result := make([]K, 0)
+
+	for i := range in {
+		for k := range in[i] {
+			if _, exists := seen[k]; !exists {
+				seen[k] = struct{}{}
+				result = append(result, k)
```
ccoVeille, line 27:
```suggestion
			_, exists := seen[k]
			if exists {
				continue
			}
			seen[k] = struct{}{}
			result = append(result, k)
```

ccoVeille: It's the logic of "leave early", it's for readability and maintainability.

First, here, it's simpler to read that when existing we skip, than reading than when no existing we have to do something.

The leave early concept is there to avoid big if branch.

Also, imagine you have an new condition in this loop

With your current code, it would lead to put an if in the if

So the code would be something like this

```
for {
    if condition1 {
         something1
         if condition2 {
               something2 
         }
    }
}
```

Here is the same code when applying leave early

```
for {
    if !condition1 {
         continue
    }
    something1
    if !condition2 {
          continue
    }
    something2 
}
```

shivamrazorpay: waaaooo this seems a very good explanation. Thank you making it soo clear

shivamrazorpay:
@samber can you review the PR ?? all the review comments are resolved.

README.md
```diff
@@ -125,9 +125,11 @@ Supported helpers for slices:
 Supported helpers for maps:
 
 - [Keys](#keys)
-- [HasKey](#HasKey)
+- [UniqKeys](#uniqKeys)
+- [HasKey](#hasKey)
 - [ValueOr](#valueor)
 - [Values](#values)
+- [UniqValues](#uniqValues)
```
ccoVeille, line 132:
```suggestion
- [UniqValues](#uniqvalues)
```

README.md
```diff
@@ -125,9 +125,11 @@ Supported helpers for slices:
 Supported helpers for maps:
 
 - [Keys](#keys)
-- [HasKey](#HasKey)
+- [UniqKeys](#uniqKeys)
+- [HasKey](#hasKey)
```
ccoVeille, line 129:
```suggestion
- [UniqKeys](#uniqkeys)
- [HasKey](#haskey)
```

ccoVeille (approved):

samber:
Please check the tests.

shivamrazorpay:
@samber Sorry my bad. There I was asserting the 3rd test and sorted the 1st test slice.
Fixed it. I checked all the tests are now working.

ccoVeille:
That's why I'm using a pattern I barely seen here.

```go
func TestWhatever(t *testing.T) {


    t.Run("one test", func(t *testing.T) {
          // …
    })

    t.Run("another test", func(t *testing.T) {
          // …
    })

   t.Run("and so on", func(t *testing.T) {
          // …
    })
}
```

ccoVeille (approved):

map.go
```diff
@@ -21,11 +42,33 @@ func HasKey[K comparable, V any](in map[K]V, key K) bool {
 
 // Values creates an array of the map values.
 // Play: https://go.dev/play/p/nnRTQkzQfF6
-func Values[K comparable, V any](in map[K]V) []V {
+func Values[K comparable, V any](in ...map[K]V) []V {
 	result := make([]V, 0, len(in))
```
d-enk, line 46: need calculate capacity 
```go
size := 0
for range in {
    size += len(in)
}
`

Comment (d-enk) [reply map.go:46]: for Keys too

File: map.go [hunk 12]
```diff
@@ -2,11 +2,32 @@ package lo
...
+			result = append(result, k)
+		}
+	}
+
+	return result
+}
+
+// UniqKeys creates an array of unique keys in the map.
+// Play: https://go.dev/play/p/TPKAb6ILdHk
+func UniqKeys[K comparable, V any](in ...map[K]V) []K {
```
d-enk, line 19: most likely more effective

```go
  seen := make(map[K]struct{})
	
	for i := range in {
		for k := range in[i] {
			seen[k] = struct{}{}
		}
	}
	
	result := make([]K, 0, len(seen))
	
	for k := range seen {
	    result = append(result, k)
	}
```

d-enk: similarly `UniqValues`

samber: with this version, the order is not guarantee

IMO, we would need a new helper (UniqKeysUnordered) ?

PR #507, opened 2024-07-26
feat: adding FromSlicePtrNotNil

samber:
Maybe we could name this helper differently.

Also, we could return a boolean when at least 1 nil value is detected and dropped. (see lo.FromAnySlice)


samber:
could be replaced by `lo.FromSlicePtr(lo.WithoutEmpty(list))`

PR #509, opened 2024-08-09
Add nil pointer protection to ternary function

LucasBN:
Allows you to use `lo.Ternary` to fallback to default values when checking if a value is nil before access an attribute:

```
lo.Ternary(example != nil, example.attribute, "fallback")
```

Previously, both arguments would be evaluated which causes a nil pointer exception. Using anonymous inlined function protects against this.

### DIFF
--- a/condition.go
+++ b/condition.go
@@ -4,10 +4,10 @@ package lo
 // Play: https://go.dev/play/p/t-D7WBL44h2
 func Ternary[T any](condition bool, ifOutput T, elseOutput T) T {
 	if condition {
-		return ifOutput
+		return func() T { return ifOutput }()
 	}
 
-	return elseOutput
+	return func() T { return elseOutput }()
 }
 
 // TernaryF is a 1 line if/else statement whose options are functions


condition.go
```diff
@@ -4,10 +4,10 @@ package lo
 // Play: https://go.dev/play/p/t-D7WBL44h2
 func Ternary[T any](condition bool, ifOutput T, elseOutput T) T {
 	if condition {
-		return ifOutput
+		return func() T { return ifOutput }()
```
ccoVeille, line 7: Please add a test for this

LucasBN: Yeah my bad this doesn't work

ccoVeille: I gently asked because I had no idea how it could be working

ccoVeille:
The only way I see to make it work would be to add something like

lo.TernaryFunc(example != nil, func() string () { return example.attribute}, func string { return"fallback"} )

samber:
Please check lo.TernaryF -> https://github.com/samber/lo?tab=readme-ov-file#ternaryf

ccoVeille:
I suggested something that already exists, haha

PR #511, opened 2024-08-11
Without By

### LINKED ISSUE #505: Proposal: Add SliceToSet, WithoutBy
It's pretty often that one needs to exclude some entries from slice by some keys

There's `lo.Without` which could be useful
But it's not always convenient cause it could be so that you have an entity and only its' **keys**
In that case it could be useful to have
```go
SliceToSet[K comparable](values []K) map[K]struct{}
```

For example:
```go
type Entity struct {
	Key string
	Value string
}

func filter(entities []Entity) []Entity {
	blackList := SliceToSet(getBlackList()) // map[string]struct{}

	return lo.Filter(entities, func(e Entity, _ int) bool {
		_, ok := blackList[e.Key]
		return !ok
	})
}
```

Yeah, one could simply return the map from `getBlackList()` or use `lo.Contains`
But to return that map `SliceToSet` could still be useful and `lo.Contains` is not always optimal
And I guess there's lots of other applications of such a function

Also seems like it could be useful adding
```go
WithoutBy[T any, K comparable](collection []T, extract func(T) K, exclude ...K) []T
```

usman1100:
# Without By

Resolves #505 

## Description

The `WithoutBy` function filters a slice of elements based on a given predicate function. It returns a new slice containing only the elements for which the predicate function returns false. 

### Filtering unverified users

```go
func isValidEmail(email string) bool {
        // Basic email validation using regular expression
        emailRegex := regexp.MustCompile(`^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$`)
        return emailRegex.MatchString(email)
}

func cleanUserList(users []User) []User {
        return WithoutBy(users, func(u User) bool { return !isValidEmail(u.Email) })
}
```

### Filtering even numbers

```go
func isEven(num int) bool {
        return num%2 == 0
}

func removeEven(numbers []int) []int {
        return WithoutBy(numbers, isEven)
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Supported intersection helpers:
 - [Union](#union)
 - [Without](#without)
 - [WithoutEmpty](#withoutempty)
+- [WithoutBy](#withoutby)
 
 Supported search helpers:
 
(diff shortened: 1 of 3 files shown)


samber:
I think you are implementing the lo.Reject helper.

PR #514, opened 2024-08-14
feat: add SliceToSet

nicklaus-dev:
see #505

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -388,6 +388,17 @@ func SliceToMap[T any, K comparable, V any](collection []T, transform func(item 
 	return Associate(collection, transform)
 }
 
+// SliceToSet returns a map with each unique element of the slice as a key.
+func SliceToSet[T comparable, Slice ~[]T](collection Slice) map[T]struct{} {
+	result := make(map[T]struct{}, len(collection))
+
+	for _, item := range collection {
+		result[item] = struct{}{}
+	}
+
+	return result
+}
+
 // Drop drops n elements from the beginning of a slice or array.
 // Play: https://go.dev/play/p/JswS7vXRJP2
 func Drop[T any, Slice ~[]T](collection Slice, n int) Slice {
--- a/slice_example_test.go
+++ b/slice_example_test.go
@@ -496,3 +496,17 @@ func ExampleIsSortedByKey() {
 
 	// Output: true
 }
+
+func ExampleSliceToSet() {
+	list := []string{"a", "b", "d"}
+
+	set := SliceToSet(list)
+	_, ok1 := set["a"]
+	_, ok2 := set["c"]
+	fmt.Printf("%v\n", ok1)
+	fmt.Printf("%v\n", ok2)
+
+	// Output:
+	// true
+	// false
+}
--- a/slice_example_test.go
+++ b/slice_example_test.go
@@ -498,15 +498,17 @@ func ExampleIsSortedByKey() {
 }
 
 func ExampleSliceToSet() {
-	list := []string{"a", "b", "d"}
+	list := []string{"a", "a", "b", "b", "d"}
 
 	set := SliceToSet(list)
 	_, ok1 := set["a"]
 	_, ok2 := set["c"]
 	fmt.Printf("%v\n", ok1)
 	fmt.Printf("%v\n", ok2)
+	fmt.Printf("%v\n", set)
 
 	// Output:
 	// true
 	// false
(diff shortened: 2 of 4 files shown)


slice_example_test.go
```diff
@@ -479,3 +481,17 @@ func ExampleIsSortedByKey() {
 
 	// Output: true
 }
+
+func ExampleSliceToSet() {
+	list := []string{"a", "b", "d"}
```
shivamrazorpay, line 486: Can you add more test , in which we have duplicate entries in the slice ?

slice.go
```diff
@@ -378,6 +378,17 @@ func SliceToMap[T any, K comparable, V any](collection []T, transform func(item
 	return Associate(collection, transform)
 }
 
+// SliceToSet returns a map with each unique element of the slice as a key.
+func SliceToSet[T comparable, Slice ~[]T](collection Slice) map[T]struct{} {
```
shivamrazorpay, line 382: Can you add this function in the Readme.

slice_example_test.go
nicklaus-dev: Sure, I fixed it! Please review it again.

slice.go
nicklaus-dev: Sure, I fixed it! Please review it again.

samber:
Thanks for your first contribution.

I'm not sure this is the best name for such a helper.

1- it is not exactly a "set" data structure, but a map (see https://github.com/emirpasic/gods)
2- it is not self-explaining until you read the function prototype (eg: SliceToMapKey)

Any takes?

Other ideas:
- make the value type dynamic: `lo.SliceToMapKey[int]([]string{"a", "b"})` -> `{"a": 0, "b": 0}`
- add a default value as parameter: `lo.SliceToXXXX([]string{"a", "b"}, 42)` -> `{"a": 42, "b": 42}`

Before merging this PR, I would like more feedback from the community 🙏

nicklaus-dev:
Hi samber,

Thank you for your advice. I understand the concern regarding the `SliceToSet` potentially causing confusion. I agree that making the method more flexible is important. In some cases, developers might use `map[T]bool` for convenience when checking for duplicates, while for performance considerations, developers might prefer `map[T]struct{}`.

To address this, `SliceToMapKey` might be a solution. This name better reflects the function's purpose and allows for greater flexibility in specifying the value type. Here is the updated implementation:

```golang
package main

import "fmt"

// SliceToMapKey returns a map with each unique element of the slice as a key and a default value.
func SliceToMapKey[T comparable, Slice ~[]T, V any](collection Slice, defaultValue V) map[T]V {
	result := make(map[T]V, len(collection))

		result[item] = defaultValue
	}

	return result
}

func main() {
	// Example usage:
	// For checking duplicates with map[T]struct{}
	result1 := SliceToMapKey([]string{"a", "b", "a"}, struct{}{})
	fmt.Printf("%v\n", result1) // map[string]struct{}{"a": {}, "b": {}}

	// For checking duplicates with map[T]bool
	result2 := SliceToMapKey([]string{"a", "b", "a"}, true)
	fmt.Printf("%v\n", result2) // map[string]bool{"a": true, "b": true}
}
```
cc @samber @shivamrazorpay @ccoVeille

senago:
I find naming `Keyify` optimal
It's a bit hard to read but the intent is clear
```go
func Keyify[T comparable](collection []T) map[T]struct{}
```

Maybe it's even worth adding
```go
func KeyifyBy[T any, K comparable](collection []T, by func(T)K) map[K]T
```
I could imagine calling it like `KeyifyBy(users, user.User.GetID) // map[user.ID]user.User`

nicklaus-dev:
> I find naming `Keyify` optimal It's a bit hard to read but the intent is clear
> 
> ```go
> func Keyify[T comparable](collection []T) map[T]struct{}
> ```
> 
> I'd stay with having `struct{}` as map value simply cause it's looks natural here
> 
> Maybe it's even worth adding
> 
> ```go
> func KeyifyBy[T any, K comparable](collection []T, by func(T)K) map[K]T
> ```
> 
> I could imagine calling it like `KeyifyBy(users, user.User.GetID) // map[user.ID]user.User`

`Keyify` is good, but `KeyifyBy` might be `SliceToMap` which has been implemented.

samber:
Thanks guy for challenging the name of this helper.

`Keyify` is very good indeed.

I'm going to make the fix and merge 👍

Issue #512, opened 2024-08-12
Allow for a custom random seed when sampling


bramvandewalle:
## Context

In one of our projects, we have a collection of objects. In case the size of a collection is bigger than a threshold, we will sample the collection. We have to sample it in a deterministic way so that the result is the same when we sample the full collection again later on.

For that, we currently do something like:

```go
func sampleObjects(objects []*pkg.Object, seed string, size int) []*pkg.Object {
	rand.Seed(getHash(seed)) //nolint: deprecated
	return lo.Samples(objects, size)
}
```

However, it is not ideal that we have to use a deprecated way of seeding `rand`.

## Request

Hence, it would be great if instead, we would be able to do something like:

```go
func sampleObjects(objects []*pkg.Object, seed string, size int) []*pkg.Object {
	r := rand.New(rand.NewSource(getHash(seed)))
	return lo.SamplesWithRand(objects, size, r)
}
```

or

```go
func sampleObjects(objects []*pkg.Object, seed string, size int) []*pkg.Object {
	hashSeed := getHash(seed)
	return lo.SamplesWithSeed(objects, size, hashSeed)
}
```

PR #516, opened 2024-08-15
feat: add SampleBy and SamplesBy

bramvandewalle:
## Why?

See https://github.com/samber/lo/issues/512

## What?

Added a `SampleBy` and `SamplesBy` that accepts an `intFunc` that is the generator of the indexes. This way, a client of the `lo` library would be able to provide their own seeded random generator.

## Example?

```go
package main

import (
	"fmt"
	"math/rand"

	"github.com/samber/lo"
)

func Example() {
	r := rand.New(rand.NewSource(42))
	collection := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
	result := lo.SampleBy(collection, r.Intn)
	results := lo.SamplesBy(collection, len(collection), r.Intn)
	fmt.Printf("Sample: %v\n", result)
	fmt.Printf("Samples: %v\n", results)
}

func main() {
	Example()
}
```


find_test.go
```diff
@@ -570,3 +583,23 @@ func TestSamples(t *testing.T) {
...
+	result2 := SamplesBy([]string{}, 3, r.Intn)
+
+	sort.Strings(result1)
+
+	is.Equal(result1, []string{"a", "b", "c"})
+	is.Equal(result2, []string{})
+
+	type myStrings []string
+	allStrings := myStrings{"", "foo", "bar"}
+	nonempty := Samples(allStrings, 2)
```
samber, line 603: SamplesBy ?

samber (changes requested):

find_test.go
bramvandewalle: Aah yes! I forgot this one. Thanks!

find.go
```diff
@@ -475,16 +475,27 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
 
 // Sample returns a random item from collection.
 func Sample[T any](collection []T) T {
+	result := SampleBy(collection, rand.IntN)
+	return result
+}
+
+// SampleBy returns a random item from collection, using intFunc as the random index generator.
+func SampleBy[T any](collection []T, intFunc func(int) int) T {
```
ccoVeille, line 483: Isn't possible to define a type here?


```suggestion
// an interesting comment to be provided here
type randomSeedFunc func(int) int

func SampleBy[T any](collection []T, intFunc randomSeedFunc) T {
```

This way both SampleBy and [SamplesBy](https://github.com/samber/lo/pull/516#discussion_r1724931129) would use this common type.

And the purpose of the initFunc would be clearer

find.go
```diff
@@ -475,16 +475,27 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
...
 }
 
 // Samples returns N random unique items from collection.
 func Samples[T any, Slice ~[]T](collection Slice, count int) Slice {
+	results := SamplesBy(collection, count, rand.IntN)
+	return results
+}
+
+// SamplesBy returns N random unique items from collection, using intFunc as the random index generator.
+func SamplesBy[T any, Slice ~[]T](collection Slice, count int, intFunc func(int) int) Slice {
```
ccoVeille, line 498:
```suggestion
func SamplesBy[T any, Slice ~[]T](collection Slice, count int, intFunc randomSeedFunc) Slice {
```

find.go
ccoVeille, line 483: Why this method doesn't call SamplesBy with count=1 and return the first entry?

bramvandewalle: Also the original implementation of the `Sample` doesn't use `Samples`. I didn't want to change any implementation of existing functions. Also... not sure if there's that much of a benefit in doing so?

bramvandewalle: But if you want me to update it... by all means and I change it!

ccoVeille: Let's wait for owner's feedback

cc @samber

bramvandewalle: I added a type `randomIntGenerator` instead.

find.go
bramvandewalle: I added a type randomIntGenerator instead.

find.go
```diff
@@ -473,18 +473,25 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
...
+type randomIntGenerator func(int) int
+
+// SampleBy returns a random item from collection, using intFunc as the random index generator.
+func SampleBy[T any](collection []T, randomIntGenerator randomIntGenerator) T {
 	size := len(collection)
 	if size == 0 {
 		return Empty[T]()
 	}
+	return collection[randomIntGenerator(size)]
+}
```
ccoVeille, line 485: I would say you have to fix the comment or the func

```suggestion
// SampleBy returns a random item from collection, using intFunc as the random index generator.
func SampleBy[T any](collection []T, intFunc randomIntGenerator) T {
	size := len(collection)
	if size == 0 {
		return Empty[T]()
	}
	return collection[intFunc(size)]
}
```


```suggestion
// SampleBy returns a random item from collection, using randomIntGenerator as the random index generator.
func SampleBy[T any](collection []T, randomIntGenerator randomIntGenerator) T {
	size := len(collection)
	if size == 0 {
		return Empty[T]()
	}
	return collection[randomIntGenerator(size)]
}
```

Please update SamplesBy accordingly, because right now the function used in it is initFunc

find.go
ccoVeille: Thanks

find.go
bramvandewalle: Apologies for that! I wanted to push it before having lunch, but didn't properly scan everything.

find.go
```diff
@@ -473,18 +473,33 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
 	return collection[l+n], nil
 }
 
+// randomIntGenerator is a function that should return a random integer in the range [0, n)
```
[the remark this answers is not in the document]
bramvandewalle: `[0, n)` and `[0, n-1]` are the same, no? `)` means an "excluding" interval.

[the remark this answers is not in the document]
bramvandewalle: `[0, n] != [0, n)`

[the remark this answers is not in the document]
ccoVeille: Oh, I have never seen this notation.

maybe, using [0,n-1] would be more commonly understood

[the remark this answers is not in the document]
bramvandewalle: Not sure if I really agree with that. It's a commonly known mathematical notation for exclusive sets... It's even defined in an ISO standard (ISO 31-11)

See https://mathresearch.utsa.edu/wiki/index.php?title=Interval_Notation#Integer_intervals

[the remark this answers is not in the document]
ccoVeille: Indeed

https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

Apparently it's also documented in the French version of the article (I'm French)
https://fr.wikipedia.org/wiki/Intervalle_(math%C3%A9matiques)#Notations

So I would tend to see you are right, it's a commonly accepted Mathematics notation, and I'm likely to agree that people who studied (advanced?) Mathematics knows it.

I did but, it was 20 years ago. So I might have forgotten.

I'm not saying you are not right with this notation, I would like to emphasize that people might not be aware of it.

Anyway, we are discussing a comment of a method. There is no need to keep talking.

I think using `[0, n-1]` might be easier and less confusing for _some people_, but I can live with `[0, n)`

Thanks for sending back to my memories of Algebra classes at college.

ccoVeille (approved):
I posted [this](https://github.com/samber/lo/pull/516#discussion_r1727969248), but feel free to merge.

find.go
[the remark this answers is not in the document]
bramvandewalle: 🙈 Woaw-ie! This got a bit out-of-hand I would say. I was just giving my 2-cents on the matter. It didn't have to become personal for me. Apologies. I updated the comment 🙃

ccoVeille (approved):

bramvandewalle:
@ccoVeille I don't have the rights to merge this. Not sure if @samber first needs to approve it?

find.go
```diff
@@ -473,18 +473,33 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
 	return collection[l+n], nil
 }
 
+// randomIntGenerator is a function that should return a random integer in the range [0, n-1]
```
samber, line 476:
```suggestion
// randomIntGenerator is a function that should return a random integer in the range [0, n)
```

this is the right math notation

find.go
```diff
@@ -473,18 +473,33 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
...
+// where n is the parameter passed to the randomIntGenerator.
+type randomIntGenerator func(int) int
```
samber, line 478:
```suggestion
type randomIntGenerator func(n int) int
```

samber (approved):

PR #517, opened 2024-08-15
do not create wait groups for slices with len less than 2

Alexfilus:
Little improvements for parallel package.


shivamrazorpay:
@Alexfilus why is the improvement helpful ?? Can you give example or any article from where you get the idea to add this change ?

Alexfilus:
In my project I am using lop.Map for processing data from db. And I don't know how many row I will get. From 0 to 20. 
And in case of zero row lop.Map is creating unnecessary waitgroup. In case of 1 row lop.Map is creating unnecessary waitgroup and goroutine. It is just useless work and memory allocations.

Alexfilus:
Not needed with new Map function

PR #519, opened 2024-08-18
Tidy up

NathanBaulch:
Just some minor tidy up - mostly spelling, grammar, newline inconsistencies, trailing whitespace, redundant types, etc.

### DIFF
--- a/retry_test.go
+++ b/retry_test.go
@@ -82,7 +82,7 @@ func TestAttemptWithDelay(t *testing.T) {
 	})
 
 	is.Equal(iter1, 1)
-	is.Greater(dur1, 0*time.Millisecond)
+	is.GreaterOrEqual(dur1, 0*time.Millisecond)
 	is.Less(dur1, 1*time.Millisecond)
 	is.Equal(err1, nil)
 	is.Equal(iter2, 6)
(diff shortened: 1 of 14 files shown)


retry_test.go
```diff
@@ -82,7 +82,7 @@ func TestAttemptWithDelay(t *testing.T) {
 	})
 
 	is.Equal(iter1, 1)
-	is.Greater(dur1, 0*time.Millisecond)
+	is.GreaterOrEqual(dur1, 0*time.Millisecond)
```
ccoVeille, line 85: How could be invalid?

I mean this test will always pass. A duration is always positive, no?

retry_test.go
```diff
@@ -187,7 +187,7 @@ func TestAttemptWhileWithDelay(t *testing.T) {
...
```
ccoVeille, line 190: Same?

retry_test.go
NathanBaulch: Could be my hardware or the fact that I'm running on Windows, but for me the first call to `AttemptWithDelay` (which succeeds immediately) returns within the minimum monotonic time period, so these tests always fail.

samber:
Wow. That much??

Me, trying to write documentation in English before the Github Copilot era:

samber:
(tests are broken)

NathanBaulch:
Were you aware that I'm Australian before posting that gif? 😛 Don't worry, we don't take ourselves too seriously!
The unit tests should be fixed now.

samber:
Ahah, sorry. I missed that little detail. 😳 🤣

PR #524, opened 2024-08-22
refactor:refactor RandomString function

pigwantacat:
reference resources：[How to generate a random string of a fixed length in Go?](https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go)
```go
func BenchmarkApproach1(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = RandomString(1<<10, lo.AlphanumericCharset)
	}
}

func BenchmarkApproach2(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = lo.RandomString(1<<10, lo.AlphanumericCharset)
	}
}
```
```bash
$ go test -benchmem -benchtime=5s -bench .
goos: windows
goarch: amd64
pkg: mytttttt
cpu: Intel(R) Core(TM) i5-10400 CPU @ 2.90GHz
BenchmarkApproach1-12            1000000              5423 ns/op            1024 B/op          1 allocs/op
BenchmarkApproach2-12             292579             20485 ns/op            5248 B/op          2 allocs/op
PASS
ok      mytttttt        12.160s
```


samber:
that's cool!

string.go
```diff
@@ -1,13 +1,14 @@
 package lo
 
 import (
+	"math"
+	"math/rand"
```
ccoVeille, line 5: You might use math/rand/v2 now

https://go.dev/blog/randv2

As you did a benchmark, you could also compare the speed.

pigwantacat: thank you
```go
func BenchmarkApproach1(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = RandomString(1<<10, lo.AlphanumericCharset)
	}
}

func BenchmarkApproach2(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = lo.RandomString(1<<10, lo.AlphanumericCharset)
	}
}

func BenchmarkApproach3(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = RandomStringV2(1<<10, lo.AlphanumericCharset)
	}
}
```
```go
$ go test -benchmem -benchtime=5s -bench .
goos: windows
goarch: amd64
pkg: mytttttt
cpu: Intel(R) Core(TM) i5-10400 CPU @ 2.90GHz
BenchmarkApproach1-12            1000000              5509 ns/op            1024 B/op          1 allocs/op
BenchmarkApproach2-12             291853             20599 ns/op            5248 B/op          2 allocs/op
BenchmarkApproach3-12            1000000              5458 ns/op            1024 B/op          1 allocs/op
PASS
ok      mytttttt        17.782s
```

ccoVeille: Quite comparable then. Unless I'm reading wrong the benchmark.

Maybe using the v2 even if it's a bit slower

pigwantacat: thanks, I have upgraded math/rand to math/rand/v2

string.go
```diff
@@ -37,12 +38,39 @@ func RandomString(size int, charset []rune) string {
 		panic("lo.RandomString: Charset parameter must not be empty")
 	}
 
-	b := make([]rune, size)
-	possibleCharactersCount := len(charset)
-	for i := range b {
-		b[i] = charset[rand.IntN(possibleCharactersCount)]
+	sb := strings.Builder{}
+	sb.Grow(size)
+	letterIdBits := int(math.Log2(float64(nearestPowerOfTwo(len(charset)))))
```
ccoVeille, line 43: You will have to explain the algorithm used here. Even by reading the article you quoted. I'm unsure I get it

string.go
```diff
@@ -37,12 +38,39 @@ func RandomString(size int, charset []rune) string {
...
+	var letterIdMask int64 = 1<<letterIdBits - 1
+	letterIdMax := 63 / letterIdBits
```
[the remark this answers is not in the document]
pigwantacat: Why do we only use 63 bits for int64? This is because the first bit indicates the signpositive or negative), and it is not random
```go
// Int64 returns a non-negative pseudo-random 63-bit integer as an int64
// from the default Source.
func Int64() int64 { return globalRand.Int64() }
```

string.go
pigwantacat: My English is not very good, I hope you can understand my explanation. For example, our random character set includes lowercase letters, uppercase letters, and digits, totaling 62 characters, which require 6 bits (62 -> 64 = 2^6) to represent. An int64 has 64 bits, but the random.Int63() method generates non-negative numbers, meaning the first bit is always fixed and not random, so there are only 63 random bits. Now with 63 available bits, one random character requires 6 bits, so one random.Int63() generated int64 can be used 10 times (63 / 6).

ccoVeille: Your English is good, no worries.

Your explanation is clear to me.

My point was to ask to comment the code with an explanation

I think you can copy-paste the one you gave here in the code.

Also, maybe one suggestion: you could add an exported method that will be in charge of producing the random number, to leave the RandomString code simpler.

pigwantacat: thank you

ccoVeille (approved):

string.go
```diff
@@ -1,13 +1,13 @@
...
 
 import (
+	"math"
+	rand2 "math/rand/v2"
 	"regexp"
 	"strings"
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/samber/lo/internal/rand"
```
samber, line 9: Please use github.com/samber/lo/internal/rand instead of math/rand/v2

string.go
```diff
@@ -25,6 +26,8 @@ var (
 	splitWordReg = regexp.MustCompile(`([a-z])([A-Z0-9])|([a-zA-Z])([0-9])|([0-9])([a-zA-Z])|([A-Z])([A-Z])([a-z])`)
 	// bearer:disable go_lang_permissive_regex_validation
 	splitNumberLetterReg = regexp.MustCompile(`([0-9])([a-zA-Z])`)
+	MaximumCapacity      = math.MaxInt>>1 + 1
```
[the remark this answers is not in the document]
pigwantacat: It's probably not necessary. It's used to prevent the value from becoming negative when it exceeds the maximum value of int.

string.go
pigwantacat: ok

PR #531, opened 2024-09-10
Added pull methods from lodash

SicParv1sMagna:
Pull methods like in lodash https://lodash.com/docs/4.17.15#pull

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -708,22 +708,9 @@ func Pull[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
 	return output
 }
 
-// PullAll is like [Pull] except that it accepts an array of values to remove
-func PullAll[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
-	output := make(Slice, 0, len(collection))
-
-	for _, item := range collection {
-		if !slices.Contains(elements, item) {
-			output = append(output, item)
-		}
-	}
-
-	return output
-}
-
-// PullAllBy is like [PullAll] except that it accepts iteratee which is invoked
+// PullAllBy is like [Pull] except that it accepts iteratee which is invoked
 // for each element of slice and values to generate the criterion by which they're compared
-func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, elements []U, iteratee func(T) U) Slice {
+func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, iteratee func(T) U, elements ...U) Slice {
 	elementCriteria := make(map[U]struct{}, len(elements))
 	for _, element := range elements {
 		elementCriteria[element] = struct{}{}
@@ -741,7 +728,7 @@ func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, elements []U, 
 
 // PullAt removes elements from slice corresponding to indexes and returns
 // an array of removed elements
-func PullAt[T any](slice []T, indexes []int) ([]T, []T) {
+func PullAt[T any](slice []T, indexes ...int) ([]T, []T) {
 	if len(slice) == 0 || len(indexes) == 0 {
 		return nil, slice
 	}
@@ -756,7 +743,7 @@ func PullAt[T any](slice []T, indexes []int) ([]T, []T) {
 	}
 
 	var removed []T
-	var output []T
+	output := make([]T, 0, len(slice)-len(indexMap))
 
 	for i, v := range slice {
 		if _, ok := indexMap[i]; ok {
--- a/slice.go
+++ b/slice.go
@@ -695,7 +695,7 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
 	return append(append(append(output, collection[:i]...), elements...), collection[i:]...)
 }
 
-// Pull removes all given values from slice using SameValueZero for equality comparisons.
+// Pull removes all given values from slice
 func Pull[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
 	output := make(Slice, 0, len(collection))
 
--- a/slice.go
+++ b/slice.go
@@ -1,6 +1,7 @@
 package lo
 
 import (
+	"slices"
 	"sort"
 
 	"github.com/samber/lo/internal/constraints"
@@ -699,15 +700,7 @@ func Pull[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
 	output := make(Slice, 0, len(collection))
 
 	for _, item := range collection {
-		shouldRemove := false
-		for _, element := range elements {
-			if item == element {
-				shouldRemove = true
-				break
-			}
-		}
-
-		if !shouldRemove {
+		if !slices.Contains(elements, item) {
 			output = append(output, item)
 		}
 	}
@@ -716,20 +709,11 @@ func Pull[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
 }
 
 // PullAll is like [Pull] except that it accepts an array of values to remove
-func PullAll[T comparable, Slice ~[]T](collection Slice, 
-...elements T) Slice {
+func PullAll[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
 	output := make(Slice, 0, len(collection))
 
 	for _, item := range collection {
-		shouldRemove := false
-		for _, element := range elements {
-			if item == element {
-				shouldRemove = true
-				break
-			}
-		}
-
-		if !shouldRemove {
+		if !slices.Contains(elements, item) {
 			output = append(output, item)
 		}
 	}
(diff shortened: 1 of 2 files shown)


samber:
What is the difference with lo.Without ?

SicParv1sMagna:
Unlike without, this method mutates array

SicParv1sMagna:
> What is the difference with lo.Without ?

Unlike without, this method mutates array

samber:
reopening, since we will add a `github.com/samber/lo/mutable` package

SicParv1sMagna:
@samber I'll wait for review! Thx

slice.go
```diff
@@ -693,3 +693,93 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+
+		if !shouldRemove {
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// This method is like Pull except that it accepts an array of values to remove
```
ccoVeille, line 718:
```suggestion
// PullAll is like [Pull] except that it accepts an array of values to remove
```

Here you would respect the convention about comment that must start with the method/function name. Also, you would link Pull with a godoc link

slice.go
```diff
@@ -693,3 +693,93 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+func PullAll[T comparable, Slice ~[]T](collection Slice, elements []T) Slice {
```
ccoVeille, line 719:
```suggestion
func PullAll[T comparable, Slice ~[]T](collection Slice, 
...elements T) Slice {
```
I would use a variadic, but then PullAll would be enough, no need for Pull.

So I would remove Pull func, and rename PullAll to Pull

slice.go
```diff
@@ -693,3 +693,93 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+
+		if !shouldRemove {
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// This method is like PullAll except that it accepts iteratee which is invoked
```
ccoVeille, line 739:
```suggestion
// PullAllBy is like [PullAll] except that it accepts iteratee which is invoked
```

slice.go
```diff
@@ -693,3 +693,93 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+	for _, item := range collection {
+		if _, shouldRemove := elementCriteria[iteratee(item)]; !shouldRemove {
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// This method removes elements from slice corresponding to indexes and returns
```
ccoVeille, line 757:
```suggestion
// PullAt removes elements from slice corresponding to indexes and returns
```

slice.go
SicParv1sMagna: Appreciate for your advice!

slice.go
```diff
@@ -693,3 +693,93 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
 }
+
+// Pull removes all given values from splice using SameValueZero for equality comparisons.
+func Pull[T comparable, Slice ~[]T](collection Slice, elements ...T) Slice {
+	output := make(Slice, 0, len(collection))
+
+	for _, item := range collection {
+		shouldRemove := false
+		for _, element := range elements {
+			if item == element {
```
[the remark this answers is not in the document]
SicParv1sMagna: Done, waiting for resolve

slice.go
```diff
@@ -693,3 +694,77 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// PullAllBy is like [PullAll] except that it accepts iteratee which is invoked
+// for each element of slice and values to generate the criterion by which they're compared
+func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, elements []U, iteratee func(T) U) Slice {
```
ccoVeille, line 726: If you invert iteratee and elements, you could use a variadic


```suggestion
func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, iteratee func(T) U, ...elements U) Slice {
```

slice.go
```diff
@@ -693,3 +694,77 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// PullAt removes elements from slice corresponding to indexes and returns
+// an array of removed elements
+func PullAt[T any](slice []T, indexes []int) ([]T, []T) {
```
ccoVeille, line 744: A variadic could be used here too

```suggestion
func PullAt[T any](slice []T, indexes ..
int) ([]T, []T) {
```

slice.go
```diff
@@ -693,3 +694,77 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+	output := make(Slice, 0, len(collection))
+
+	for _, item := range collection {
+		if !slices.Contains(elements, item) {
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
```
ccoVeille, line 722: You no longer need it

```suggestion
```

slice.go
SicParv1sMagna: Done

slice.go
SicParv1sMagna: Done

slice.go
SicParv1sMagna: Done

SicParv1sMagna:
@ccoVeille thank you for your review

ccoVeille:
> @ccoVeille thank you for your review

You are welcome

slice.go
```diff
@@ -693,3 +694,64 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+// PullAt removes elements from slice corresponding to indexes and returns
+// an array of removed elements
+func PullAt[T any](slice []T, indexes ...int) ([]T, []T) {
+	if len(slice) == 0 || len(indexes) == 0 {
+		return nil, slice
+	}
+
+	sort.Sort(sort.Reverse(sort.IntSlice(indexes)))
+
+	indexMap := make(map[int]struct{}, len(indexes))
```
ccoVeille, line 738: Please check and test the behavior if you pass multiple times the same index in `indexes`

I'm worried because of the code a few lines below

```go
output := make([]T, 0, len(slice)-len(indexMap))
```

Imagine a slice with 1 item and pass 0,1,2 as indexes, or 0,0,0

One solution could be to call slices.Compact on indexes before creating indexMap

slice.go
```diff
@@ -693,3 +694,64 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+
+	return output
+}
+
+// PullAt removes elements from slice corresponding to indexes and returns
+// an array of removed elements
+func PullAt[T any](slice []T, indexes ...int) ([]T, []T) {
+	if len(slice) == 0 || len(indexes) == 0 {
+		return nil, slice
+	}
```
ccoVeille, line 734: What happens if you pass a negative index?

Imagine passing -1, -2

Is the code ok?

Maybe using uint instead of int

slice.go
```diff
@@ -693,3 +694,64 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
 
 	return append(append(append(output, collection[:i]...), elements...), collection[i:]...)
 }
+
+// Pull removes all given values from slice using SameValueZero for equality comparisons.
```
ccoVeille, line 698:
```suggestion
// Pull removes all given values from slice using SameValueZero for equality comparisons.
```

I'm sorry but you are not using SameValueZero to check the item to filter. It looks like a copy pasta from Splice

slice.go
SicParv1sMagna: Its cool that you have noticed that, I really hadn't thought about that case. I'll tests for this case, but we don't need to use uint because it works correct even with negative indexes

slice.go
SicParv1sMagna: Thank you for your comment, I has checked first case and described it in tests, that was working correctly.
Second case working correctly too, added test cases for both

slice.go
SicParv1sMagna: Done

ccoVeille (approved):

SicParv1sMagna:
@samber can you merge branch?

slice.go
```diff
@@ -693,3 +694,64 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+	for _, item := range collection {
+		if _, shouldRemove := elementCriteria[iteratee(item)]; !shouldRemove {
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// PullAt removes elements from slice corresponding to indexes and returns
```
samber, line 729: duplicated with WithoutNth

slice.go
```diff
@@ -693,3 +694,64 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+			output = append(output, item)
+		}
+	}
+
+	return output
+}
+
+// PullAllBy is like [Pull] except that it accepts iteratee which is invoked
+// for each element of slice and values to generate the criterion by which they're compared
+func PullAllBy[T any, U comparable, Slice ~[]T](collection Slice, iteratee func(T) U, elements ...U) Slice {
```
samber, line 713: should be renamed to `WithoutBy` and moved to intersect.go

samber (changes requested):

samber:
Closing, since helpers have been implemented in other PRs.

PR #534, opened 2024-09-14
Remove redundant params of Filter, Map, FilterMap and FlatMap

miranquil:
There's no need to declare an `int index` without any usage. 
Since these internal functions won't receive the slice itself, the index won't be able to be used to obtain the items.

### DIFF
--- a/map_benchmark_test.go
+++ b/map_benchmark_test.go
@@ -6,8 +6,9 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


samber:
This a major breaking change.

IMO, we should have 2 helpers:
- `lo.Map(collection, func(T) U { ... })`
- `lo.MapI(collection, func(T, int) U { ... })`

See #80

PR #537, opened 2024-09-18
Update README.md

guyareco2:
Makes more sense to check in the Some an example when some of the values are present but not every one of them


samber:
sure

PR #541, opened 2024-09-29
docs: fix example README, add playground (FromSlicePtrOr)

snamiki1212:
@samber SSIA. Thank you for nice package!
To update README and add playground for `FromSlicePtrOr`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2759,7 +2759,7 @@ Returns a slice with the pointer values or the fallback value.
 str1 := "hello"
 str2 := "world"
 
-ptr := lo.FromSlicePtrOr[string]([]*string{&str1, &str2, nil}, "fallback value")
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2759,10 +2759,12 @@ Returns a slice with the pointer values or the fallback value.
 str1 := "hello"
 str2 := "world"
 
-ptr := lo.FromSlicePtrOr[string]([]*string{&str1, &str2, "fallback value"})
+ptr := lo.FromSlicePtrOr[string]([]*string{&str1, &str2, nil}, "fallback value")
```
samber, line 2762:
```suggestion
ptr := lo.FromSlicePtrOr([]*string{&str1, nil, &str2}, "fallback value")
```

PR #547, opened 2024-10-31
feat: add In function

thedevsaddam:
**Introduce `In` function for element existence check in a collection**

Checks if an element exists in the collection. It returns true if the element is found; otherwise, false.

```go
result := lo.In([]int{1, 2, 3, 4, 5}, 4)
// true

result = lo.In([]string{"a", "b"}, "c")
// false

```

> Note: This can be done using existing `lo.Count([]int{1, 5, 1}, 1)>0` or using `lo.CountBy`

**The `In` function offers several advantages:**
- Readability and Intent:  The `In` function explicitly states its purpose of checking for element existence, making the code easier to understand.
- Performance: The`In` function returns immediately upon finding the first match, while the Count function iterates through the entire collection even if it finds a match early. This can lead to unnecessary iterations, especially in large collections
- Return Value: The `In` function returns a boolean, which is a straightforward answer to the question of whether the element exists. In contrast, Count returns a count, which is more information than needed for a simple existence check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ Supported helpers for slices:
 - [IsSorted](#issorted)
 - [IsSortedByKey](#issortedbykey)
 - [Splice](#Splice)
+- [In](#In)
 
 Supported helpers for maps:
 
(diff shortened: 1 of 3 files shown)


SnowOnion:
`lo` contains `lo.Contains` already. ^_^
[lo.Contains](https://pkg.go.dev/github.com/samber/lo#Contains) :: `func[T comparable](collection []T, element T) bool`

Since go1.21, stdlib `slices` is added and it contains `Contains`, too. 
[slices.Contains](https://pkg.go.dev/slices#Contains) :: `func[S ~[]E, E comparable](s S, v E) bool`


[2024-11-01 09 21 36](https://godoogle.sonion.xyz/search?q=%5BT+comparable%5Dfunc%28collection+%5B%5DT%2C+needle+T%29+bool)

https://godoogle.sonion.xyz/search?q=%5BT+comparable%5Dfunc%28collection+%5B%5DT%2C+needle+T%29+bool

thedevsaddam:
Thank you for the feedback! Given that similar functionality already exists, I'll go ahead and close this PR. I appreciate the insights!

Issue #224, opened 2022-09-24
Implement Min/Max as variadic functions


dergus:
Having`Min`/`Max` functions take a slice as an argument is a bit inconvenient when you have just two variables because then you have to create an array out of them just for calling Min/Max functions. 
And with variadic version you still could provide a slice if you wished by using  the `...` operator.

rkperes:
Hi @dergus,

For the current available functions, that'd be a breaking change. See an example [here](https://go.dev/play/p/GxcmDQx0nci). While a variadic parameter may take a slice, you're still required to add `...` after the slice argument, making it a breaking change.

We could still consider a second set of functions with the variadic parameters. @samber, if you think that makes sense, I would like to work on that as part of hacktoberfest.

dergus:
Hi, I see
That's also an option. I have a pr for this but I can reimplement as a separate set of functions

samber:
Hi @dergus & @rkperes 

I agree, but I am reluctant to any breaking change. I will let this issue+PR open until v2.0.0.

dashjay:
I implemented a version in more convenient way, I do not agree to change the interface, because the interface of the base library once opened to change will cause a lot of dissatisfaction.

```golang
func MaxN[T constraints.Ordered](items ...T) T {
	return Max(items)
}

func MinN[T constraints.Ordered](items ...T) T {
	return Min(items)
}
```

If you @samber looks good,  please approve it.

K4L1Ma:
In Go 1.21, built-in min and max functions were introduced, allowing developers to determine the minimum and maximum values among a set of comparable elements without the need for external libraries. 

I don't think this needs to be implemented as the language already provides a built-in variadic approach

https://tip.golang.org/ref/spec#Min_and_max

PR #557, opened 2024-11-12
feat(find): implement maxn & minn for variadic count items

dashjay:
After reviewing code in issue #224,I implemented a version in more convenient way, I do not agree to change the interface, because the interface of the base library once opened to change will cause a lot of dissatisfaction.

```go
func MaxN[T constraints.Ordered](items ...T) T {
	return Max(items)
}

func MinN[T constraints.Ordered](items ...T) T {
	return Min(items)
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,10 +223,12 @@ Supported search helpers:
 - [FindDuplicates](#findduplicates)
 - [FindDuplicatesBy](#findduplicatesby)
 - [Min](#min)
+- [MinN](#minn)
 - [MinBy](#minby)
 - [Earliest](#earliest)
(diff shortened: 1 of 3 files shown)


dashjay:
#224

dashjay:
@samber  Thank you for open such a nice package!
PTLA~

samber:
Hi @dashjay 

I think we will eventually make a breaking change in v2 to convert these helpers to varargs.

In that time, creating a new helper that would also need to be deprecated before v2 seems like a bad idea to me, as it would result in a double breaking change.

I'm closing the issue, but feel free to open it again for more discussions.

Issue #330, opened 2023-03-20
Curried map, reduce and filter for partial application 


Southclaws:
I make use of partial application quite a lot, especially with lodash/fp, it's handy for building ad-hoc easily testable transformers.

Would be nice to see some added here! One repo that's growing out of #82 is `dt` where I'm dropping all these useful functions but it would be great to have it all in one place.

https://github.com/Southclaws/dt

An example of partial application would be:

```go
func Map[T any, R any](iteratee func(T) R) func([]T) []R {
	return func(collection []T) []R {
		result := make([]R, len(collection))

		for i, item := range collection {
			result[i] = iteratee(item)
		}

		return result
	}
}
```

so the argument order is flipped and separated into two separate function calls. This allows you to "build" a "mapper": `mapper := Map(someFunc)` then re-use it later: `updated := mapper(list)`

itroot:
It is also possible to use something like it as well:
```go

// uncurry helps to use lo.Map without providing an index to iteratee
func uncurry[T any, R any](f func(T) R) func(T, int) R {
	return func(t T, _ int) R {
		return f(t)
	}
}
```

To transform a mapper with type `func (T) R` into `func(T, int) R` type. [An example on go.dev](https://go.dev/play/p/e6Q-uT7yOg8). Please ping me if anyone wants a PR for it. Not sure it will be accepted though. Still, qute handy and could be useful in many other places in lo.

![image](https://github.com/user-attachments/assets/dc71872e-dfcc-45df-ad4b-b6f698f6feee)

PR #565, opened 2024-11-28
feat(func): add `FuncWithIndex` and `FuncWithIndexNoReturn`

quantumsheep:
Hey! I regularly write some functions that I pass to `lo.Map` and it's a hassle to have to pass the index every time.

Currently:
```go
lo.Map(slice, func (v string, _ int) string {
	return doThing(v)
})
```

With the PR:
```go
lo.Map(slice, lo.FuncWithIndex(doThing))
```

I'm not sure about the naming, it could simply be `WithIndex`.

I hope you like the idea.

### DIFF
--- a/func.go
+++ b/func.go
@@ -39,3 +39,15 @@ func Partial5[T1, T2, T3, T4, T5, T6, R any](f func(T1, T2, T3, T4, T5, T6) R, a
 		return f(arg1, t2, t3, t4, t5, t6)
 	}
 }
+
+func FuncWithIndexNoReturn[T any](f func(T)) func(T, int) {
+	return func(v T, _ int) {
+		f(v)
+	}
+}
+
+func FuncWithIndex[T any, U any](f func(T) U) func(T, int) U {
+	return func(v T, _ int) U {
+		return f(v)
+	}
+}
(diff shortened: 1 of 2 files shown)


itroot:
Wow. also faced that issue, and made a quite similar solution:
- https://github.com/samber/lo/issues/330#issuecomment-2507770542

quantumsheep:
> Wow. also faced that issue, and made a quite similar solution:
> 
>     * [Curried map, reduce and filter for partial application  #330 (comment)](https://github.com/samber/lo/issues/330#issuecomment-2507770542)

Nice to see I'm not the only one wanting a feature like this :)

samber:
I agree.

But I would vote for duplicating `lo.Map` helper:
- `lo.Map`
- `lo.MapI`

quantumsheep:
@samber `lo.MapI` would be the version with or without index?

If it's the one with index, I wonder if that's a good idea as it will break many projects. Or it will require a major update.

What do you think?

samber:
Yes 😅

I would break a large ecosystem.
We would need to group all that changes into a single v2 major update.

I'm just sharing some ideas. Not telling we are going to do it right now 😇

quantumsheep:
Then I really prefer this option! I'm not sure how people use `lo.Map` right now but I guess the index argument is not that much used.

Is there a place where contributors can discuss the changes for v2? I would like to help 😄

samber:
@quantumsheep see #579

PR #566, opened 2024-12-03
feat: add `Product` and `ProductBy` functions

JohnDevitt:
Currently playing around with Go on Advent of Code and found that a Product or ProductBy function would be useful, and fit in well with the preexisting Sum and SumBy functions

Used as such:

```
list := []int{1, 2, 3, 4, 5}
product := lo.Product(list)
// 120
```

```
strings := []string{"foo", "bar"}
product := lo.ProductBy(strings, func(item string) int {
    return len(item)
})
// 9
```

Hope its useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ Supported math helpers:
 - [Clamp](#clamp)
 - [Sum](#sum)
 - [SumBy](#sumby)
+- [Product](#product)
+- [ProductBy](#productby)
 - [Mean](#mean)
 - [MeanBy](#meanby)
 
@@ -787,7 +789,6 @@ l := lo.DropByIndex([]int{0, 1, 2, 3, 4, 5}, 2, 4, -1)
 
 [[play](https://go.dev/play/p/JswS7vXRJP2)]
 
-
 ### Reject
 
 The opposite of Filter, this method returns the elements of collection that predicate does not return truthy for.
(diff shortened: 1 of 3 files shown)


artemnesterenko:
I think "Multiply" could be a better name, because "Product" can be related to Cartesian Product which is also a nice-to-have function

JohnDevitt:
> I think "Multiply" could be a better name, because "Product" can be related to Cartesian Product which is also a nice-to-have function

I'd argue `Product` is the proper name for the actual functionality here. Multiply in mind in the product of 2 numbers, rather than `n` numbers. Even in the lodash documentation their [multiply function](https://lodash.com/docs/4.17.15#multiply) says it "Returns the product" (Possibly a bad naming choice they don't want to introduce a breaking change for)

How would you feel about `Product` for these functions and a follow-up PR with a `CartesianProduct` or `CrossJoin` (I believe this is used in other libraries) function? Both seem a bit clearer to me and better describe their respective operation

JohnDevitt:
> I think "Multiply" could be a better name, because "Product" can be related to Cartesian Product which is also a nice-to-have function

> How would you feel about `Product` for these functions and a follow-up PR with a `CartesianProduct` or `CrossJoin` (I believe this is used in other libraries) function?

https://github.com/samber/lo/pull/567

samber:
I vote for `product` 😇

samber (approved):

samber:
`itertools` libs in python and rust use `product`

PR #572, opened 2025-01-07
fix: order of GroupBy and PartitionBy

### LINKED ISSUE #510: lop.PartitionBy does not guarantee the same order
I have tested the example of[ lop.PartitionBy on README.md](https://github.com/samber/lo?tab=readme-ov-file#partitionby)
It keep changed the order

liyishuai:
Fix #510


samber (approved):

samber:
Thanks for the PR !

I think we need to improve this sub-package !

PR #475, opened 2024-06-27
feat: speed up loops by reducing allocations

samber:

PR #587, opened 2025-02-05
Perf(slice): Optimize iteration function parameters and add test cases

shuqingzai:
- Modify the iteration function parameters in the `UniqueBy`, `GroupByMap`, and `Keyify` functions to use index access for collection elements. See: #428 #475 
- Optimize parameter naming in the `GroupByMap` function to improve code readability. See: #443 
- Add test cases for the `GroupByMap` function to cover different types of inputs. See: #443


samber:
Thanks!

PR #588, opened 2025-02-06
Add push, pop, shift, & unshift functions.

FGasper:
This allows code like the following:
```
foo := []int{1, 2, 3}
lo.Push(&foo, 4)
```
… which is a bit more concise than Go’s familiar `foo = append(foo, 4)` idiom.

If the maintainers agree that these are good additions, I’ll add tests before merging.

Thanks!

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -742,3 +742,62 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
 
 	return append(append(append(output, collection[:i]...), elements...), collection[i:]...)
 }
+
+func verifyNonEmpty[T any](slice []T) {
+	if len(slice) == 0 {
+		panic("Cannot remove element from empty slice!")
+	}
+}
+
+// Push adds the given elements to the end of the slice
+// and returns the slice's new length.
+//
+// Equivalent to, but more concise, than:
+//
+//	collection = append(collection, elements...)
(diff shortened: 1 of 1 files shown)


samber:
Please move this work to mutable/ directory.

slice.go
```diff
@@ -742,3 +742,88 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
...
+	return el
+}
+
+// Unshift adds the given elements to the start of the slice
+// and returns the slice's new length.
+//
+// Equivalent to, but more concise than:
+//
+//	collection = append(elements, collection...)
+func Unshift[T any, Slice ~[]T](collection *Slice, elements ...T) int {
```
samber, line 822: In respect with go semantic, i would replace Push/unshift by Append/Preppend.

slice.go
```diff
@@ -742,3 +742,88 @@ func Splice[T any, Slice ~[]T](collection Slice, i int, elements ...T) Slice {
 
 	return append(append(append(output, collection[:i]...), elements...), collection[i:]...)
 }
+
+func verifyNonEmpty[T any](slice []T) {
+	if len(slice) == 0 {
+		panic("Cannot remove element from empty slice!")
```
samber, line 748: I don't really like calls to panic in such utilitary lib, any other idea?

sejin-P: I think it should return bool value which indicates whether it is valid or not.

PR #589, opened 2025-02-07
Add ElementsMatch, ElementsMatchBy

senago:


senago:
I'm actually not sure If we should count the elements. So is done in testify package: [ElementsMatch](https://pkg.go.dev/github.com/stretchr/testify/require#ElementsMatch). That's the intent - to have the same familiar semantics. Though personally I don't find this counting really helpful.

Issue #605, opened 2025-03-02
Documentation for Ternary function - could suggest you use TernaryF instead to avoid the nil-pointer trap.


peterhoward42:
The documentation for the `TernaryF()` function nicely explains how it is a good way to avoid the nil pointer error that occurs when you use the non-F variant like this `lo.Ternary(example != nil, example.attribute, "fallback")
`.

BUT the documentation for the non-F variant (which I imagine most people discover and use first) does NOT tell you that the implementation evaluates both branches regardless, and thus gives rise to the nil pointer error.

I've used it hundreds of times and never hit the problem until today. I went straight to the documentation and saw nothing to concern me, so then searched for known bugs in this issue tracker. I came across a closed ticket - which alerted me to the function variant, and solved my problem. But a small documentation change would have saved me a lot of time.

Great work by the way. Thanks for the package. I find this function and the generic `Map()` function in particular to be invaluable.

peterhoward42:
See corresponding PR
https://github.com/samber/lo/pull/606

PR #606, opened 2025-03-02
fix: Issue 605 (README for Ternary()

peterhoward42:
Tiny addition suggested to warn about unexpected nil-pointer trap.

See https://github.com/samber/lo/issues/605


peterhoward42:
Done

PR #608, opened 2025-03-06
feat: safer version sync.WaitGroup

yuweiweiouo:
# Summary

This PR introduces a safer version of Go's standard `sync.WaitGroup` with improved error handling and context support.

# Changes

- Implemented a wrapper around `sync.WaitGroup` that captures and returns errors
- Added panic recovery to prevent goroutine crashes from affecting the entire application
- Added context support with `GoWithContext` and `GoWithContextError` methods
- Ensured proper synchronization with mutex protection for error collection

# Motivation

The standard `sync.WaitGroup` requires manual management of `Add` calls, which can lead to errors if not handled properly. This implementation simplifies concurrency management by:

- Automatically managing `WaitGroup.Add`, reducing manual overhead
- Preventing panics from crashing the application
- Allowing errors to be collected and handled properly
- Supporting context cancellation for better resource management

All functionality is thoroughly tested and ready for production use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # lo - Iterate over slices, maps, channels...
 
 [![tag](https://img.shields.io/github/tag/samber/lo.svg)](https://github.com/samber/lo/releases)
@@ -311,6 +310,7 @@ Concurrency helpers:
 - [Transaction](#transaction)
 - [WaitFor](#waitfor)
 - [WaitForWithContext](#waitforwithcontext)
+- [WaitGroup](#waitgroup)
 
 Error handling:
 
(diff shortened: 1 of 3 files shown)


ccoVeille:
Naive question: why this when https://pkg.go.dev/golang.org/x/sync@v0.12.0/errgroup exists ?

yuweiweiouo:
> Naive question: why this when https://pkg.go.dev/golang.org/x/sync@v0.12.0/errgroup exists ?
This isn't meant to replace errgroup, but rather just a small utility to simplify the common pattern of using sync.WaitGroup with goroutines.

With a regular WaitGroup, you always need to remember to call Add() before launching a goroutine and Done() inside it, which can be a bit repetitive. This just wraps that pattern to make it a little more ergonomic.

yuweiweiouo:
> Naive question: why this when https://pkg.go.dev/golang.org/x/sync@v0.12.0/errgroup exists ?

Oh, I didn't realize there was already a tool like this—thank you for pointing it out! errgroup actually achieves the same goal I had in mind, simplifying the common pattern of using sync.WaitGroup with goroutines.😊

PR #609, opened 2025-03-09
feat: Added time constants

liorkesten:
## PR Description

Added time constants for Day, Week, Month, and Year to ensure consistency and simplify access, avoiding inline calculations like 24 * time.Hour.

### DIFF
--- a/time.go
+++ b/time.go
@@ -2,6 +2,13 @@ package lo
 
 import "time"
 
+const (
+	Day   = 24 * time.Hour
+	Week  = 7 * Day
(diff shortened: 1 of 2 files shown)


samber:
hey @liorkesten 

There is a good reason if go does not bring those constants already.

1 day is not always 24 hours
1 month is not always 30 days
1 year is not always 365 days

The stdlib comes with `t.AddDate(...)`, but i think there are some libraries for accurate computations.

Feel free to argue until I close the PR.

PR #615, opened 2025-03-30
Add FilterKeys for map

rasha108bik:
Added FilterKeys transforms a map into a slice based on predicate returns truthy for specific elements.

### DIFF
--- a/find.go
+++ b/find.go
@@ -107,6 +107,19 @@ func FindKeyBy[K comparable, V any](object map[K]V, predicate func(key K, value 
 	return Empty[K](), false
 }
 
+// FindKeysBy returns the values of the elements predicate returns truthy for.
+func FindKeysBy[K comparable, V any](object map[K]V, predicate func(key K, value V) bool) []V {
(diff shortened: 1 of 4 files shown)


samber:
Hey @rasha108bik 

I wonder if we should rename this method to `FilterKeys` ?

Since `FindXXX` predicates only return a boolean and the final outputs are not slice, it would make sense.

WDYT ?

rasha108bik:
Hey @samber
Thanks, that's a good idea! I fixed it, look again.

rasha108bik:
Hey @samber
Thanks, that's a good idea! I fixed it, look again.

rasha108bik:
> Hey @rasha108bik
> 
> I wonder if we should rename this method to `FilterKeys` ?
> 
> Since `FindXXX` predicates only return a boolean and the final outputs are not slice, it would make sense.
> 
> WDYT ?

Hey @samber
Thanks, that's a good idea! I fixed it, look again.

PR #621, opened 2025-04-17
feat: add slice converter

cold-bin:
👋 Hi there!

I'd like to propose a feature enhancement for the `lo` library. In my daily development work, I frequently encounter scenarios where I need to:

- Convert between different slice types
- Transform slice elements from one type to another
- Extract specific fields from a slice of structs (e.g., extracting ID fields from a slice of database model structs)

I've noticed that `lo` currently doesn't provide utility functions for these operations. Given the general usefulness of such functionality, I would love to contribute to implementing this feature.

💡 This new utility function would help developers:
- Reduce boilerplate code
- Improve code readability
- Simplify common slice operations

I'm excited to discuss this idea with the community!

Best regards


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/samber/lo
 
-go 1.18
+go 1.23.0
+
+toolchain go1.24.1
```
Sianao, line 5: Why do we need to add the toolchain and upgrade to a newer version? Is the older version missing support for this function?

cold-bin: ok, I updated the pr

PR #624, opened 2025-04-27
Added 5 more utility functions for errors and type manipulation

Akalanka47000:
Usage and purpose for the following functions are documented within the changes

Errors:
- Ok
- OkOr

Type Manipulation:
- Cast
- CastJSON
- FromBytes


type_manipulation.go
```diff
@@ -187,3 +190,27 @@ func CoalesceMapOrEmpty[K comparable, V any](v ...map[K]V) map[K]V {
...
+	bytes, _ := json.Marshal(val)
+	return FromBytes[T](bytes)
+}
+
+// Converts a byte array to a given type. Ignores any errors.
+func FromBytes[T any](bytes []byte) T {
+	var v T
+	_ = json.Unmarshal(bytes, &v)
+	return v
+}
```
samber, line 216: I think it's not a good idea to bring such helper in `lo`. Also the naming is really bad.

type_manipulation.go
```diff
@@ -187,3 +190,27 @@ func CoalesceMapOrEmpty[K comparable, V any](v ...map[K]V) map[K]V {
 	}
 	return map[K]V{}
 }
+
+// Converts any type to a given type. If conversion fails, it returns the zero value of the given type. This is a type-safe version of type assertion.
+//
+// Cast enforces strict type assertion, for example trying to convert a number of 10 to float64 will return 0.0 instead of 10.0.
+func Cast[T any](val any) T {
```
samber, line 197: you can already do `a := val.(A)` (skipping the "ok") 

so what is the use case for such helper?

errors.go
```diff
@@ -352,3 +352,19 @@ func ErrorsAs[T error](err error) (T, bool) {
 	ok := errors.As(err, &t)
 	return t, ok
 }
+
+// Ok returns the value and ignores the error. Use with caution and only when you don't care about the error.
```
samber, line 356: I like the idea of both `Ok` and `OkOr`, but the method name does not seems "self-explaining" to me. Can you suggest other names ?

Anyway, we need to rename `Ok` to `OkOrEmpty` to match the convention of this library.

samber: `OrEmpty`/ `OrElse` ?

samber: @yuweiweiouo suggested `EmptyOnError` in #495

PR #645, opened 2025-06-26
chore: added a nixos flake file

lanathlor:
Hello. I'm using NixOS, and flakes like this are a good way to list dependencies for a project and make other NixOS users able to work on the repo

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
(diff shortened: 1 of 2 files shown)


samber:
I'm ok with devcontainer, but flakes seem too niche to me. Just run `make install` instead.

lanathlor:
Hello. Thank you for the answer
In case some nixos user and up here, I'll keep maintaining the flake in my fork. 
Do you want me to create a merge request for devconainters ?
I dont have the go + gnumake toolchain installed globally, its the flake.nix that is taking care of that. So just running `make install` will not work for some nixos users. But the devcontainers can be a solution

samber:
Yes, you can create the devcontainer config ! Thanks.

PR #646, opened 2025-06-26
feat: add Curry function

lanathlor:
Hello.
This commit introduces a set of `CurryN` functions to transform standard multi-argument functions into curried form. Each `CurryN` takes a function with N parameters and returns a nested chain of unary functions that can be called one argument at a time.

Example:
  Curry3(func(x, y, z int) int { return x + y + z }) → func(int) func(int) func(int) int

These utilities follow the functional programming style established in the library and can be found on other TS libs like [lodash](https://www.geeksforgeeks.org/javascript/lodash-_-curry-method/) or [ramda](https://ramdajs.com/docs/#curry).

Here we are limited by go and we need to implement it manually for all arity (or at least i didnt found a workaround)


BetaMedina:
Hello, I’d like to share my thoughts here. The approach taken has some flaws that could be problematic. Don’t you find it strange to have to implement repeated methods just to make your functionality work?
How would it behave if I wanted a Curry10, for example?

From my perspective, the fact that methods need to be repeated starts to become an issue…

Couldn’t using Any or perhaps an approach with Reflect make it more agnostic?

ghostsquad:
> Hello, I’d like to share my thoughts here. The approach taken has some flaws that could be problematic. Don’t you find it strange to have to implement repeated methods just to make your functionality work? How would it behave if I wanted a Curry10, for example?
> 
> From my perspective, the fact that methods need to be repeated starts to become an issue…
> 
> Couldn’t using Any or perhaps an approach with Reflect make it more agnostic?

A function that takes 10 arguments needs to be refactored to take fewer arguments.

lanathlor:
Sorry for the delay, im using my branch on my projects and am happy with this setup so i wasnt checking this PR

Agreed with @ghostsquad. Also, im using this function in hot paths where reflection would cost too much pref for my test. But i agree with the design flow. I would like to find a better solution, like what you can do in typescript, but did not found it

PR #653, opened 2025-07-28
Feature/intersect by

ghosx:
feat: Add IntersectBy function and tests, supporting intersection by custom key sets


NathanBaulch:
It's happening again @samber - CI broken due to linting issue :grimacing:

PR #655, opened 2025-08-03
V2: improve data-structure modification functions

sr9000:
- reorganize data-structure modification functions into separate sub-packages
- using same rules for functions naming across all sub-packages
- using generator for repetative function templates

brief overview of a packages

1. loany - funcs that operate on `any` or converts from/to `[]any` or `map[K]any`
2. lofn - functional-style helpers, especially `Partial` ones (https://en.wikipedia.org/wiki/Partial_application)
3. lomap - non-destructive operations on maps
4. loset - specialisation of map that uses only keys `map[K]struct{}`
5. loslice - non-destructive operations on slices
6. lotup - heavily code-generated sub-package to operate with `Tuple<N>`
7. lotype - tiny sub-package to have lambda-caster on ints/strings.
8. mutmap - destructive operations on maps
9. mutslice - destructive operations on slices

this PR is draft only and a lot of work is waiting (especially tests) and before to adopt that changes some questions must be answered:
1. is this PR ever can be approved or separate package is preferable?
2. is this PR become version 2 or changes can be live along with original functions?


sr9000:
@samber Hi! I'm looking for the best way to delivery suggested improvements into lodash codebase :rocket: 

What can I do to make it possible?

PR #669, opened 2025-09-19
lint: Apply testifylint linter recommendations

NathanBaulch:
Made some minor tweaks to the unit tests based on recommendations from the `testifylint` golangci-lint linter.
The most common issue was reversed `expected` and `actual` arguments, which can cause confusing failure messages.


samber:
Thanks !

I'm going to add a golangci-lint config soon.

PR #671, opened 2025-09-23
lint: pin golangci-lint version

NathanBaulch:
Pin the golangci-lint version so that contributors aren't blocked by unrelated CI issues if a new version happens to be released. This PR pins it to the previous version `v2.4` which this codebase passes. Version `v2.5` was released a couple of days ago with new issues to be addressed.


samber:
I would rather keep the latest golangci-lint version.

Thanks for the suggestion

NathanBaulch:
No worries. A lot of OSS project maintainers wont even look at PRs unless all the CI checks pass but that's not the case here.

PR #672, opened 2025-09-24
feat: new iter package

NathanBaulch:
I finally got around to implementing iter support, as discussed last year in #439. This turned into a pretty big task, with over 160 new helper functions based on existing slice oriented helpers in the `iter` sub-package! :sparkles: 

Implementation notes:
1. Requires Go 1.23 - might need to be part of a v2 release. Alternatively it could be a sub-module.
2. I've followed the `I` suffix convention used in the `mutable` package for helper callbacks that include an index parameter.
3. Variadic parameters have been left as is - converting to sequences made helpers awkward to use.
4. `Difference`, `FilterReject`, `Unzip*` not supported - no reasonable way to fan-out to multiple sequences.
5. `Reverse`, `ReduceRight` not supported - would need to buffer entire sequence.
6. `Sample*` not supported - sequence length is unknown and no way to seek randomly forwards/backwards.
7. `DropByIndex`, `Subset`, `*Nth*`, `Splice` don't support negative indexes.
8. `GroupBy`, `PartitionBy`, `Mode` return slices - they already have to scan the full input sequence.
9. `Chunk` returns a sequence of slices - feels awkward returning a sequence of sequences. Can we reuse the allocated slice?
10. `FindUniques*` is implemented but needs to fully iterate the input sequence before yielding anything - not sure if reasonable.
11. `FindDuplicates*` yields items in order of the first duplicate - this order differs slightly from the slice version. 
12. `Fill` is implemented but the name feels misleading - there's no pre-allocated memory being "filled".
13. `DropByIndex`, `WithoutNth` are identical - the only difference between the slice versions is `DropByIndex` supports negative indexes.
14. `DropRight` allocates an internal buffer slice of size n.
15. `DropRightWhile` allocates an internal buffer slice that grows to the longest sub-sequence of items for which the predicate returns true.
16. Map helpers use `iter.Seq2` and do not use `lo.Entry` - more idiomatic.
17. `Entries`, `ToPairs`, `FromEntries`, `FromPairs` accept multiple variadic inputs similar to `Keys` and `Values` - otherwise these helpers wouldn't offer any benefit over `maps.All` and `maps.Collect`.
18. Users will need to be careful about passing infinite sequences to helpers like `lo.Sum` - will block forever.
19. Passing a nil sequence panics - this is consistent with `slices.Collect`.
20. I've used signed `int` or `constraints.Integer` everywhere even if negative arguments are not supported.


NathanBaulch:
I've had a lot of issues getting unit tests passing reliably due to timing dependencies in the channel/concurrency/retry tests - nothing to do with the specific changes introduced in this PR.
FWIW, I managed to get them passing in my fork after a few manual retries: https://github.com/NathanBaulch/lo/actions/runs/17966454606

samber:
Hey @NathanBaulch 

A big thank you for this work. 👏

---

- `1.` Instead of supporting only go >= 1.23, what about a directive similar to the [internal/rand/ package](https://github.com/samber/lo/tree/master/internal/rand) ? I maintain this library as backward compatible as possible, for broad usage and LTS.
- `3.` I agree.
- `5. + 6.` Since you added `GroupBy`+`PartitionBy`+`Mode`, i don't feel uncomfortable adding such helpers, especially if this is properly documented.
- `11.` It does not feel safe to me to reuse the slice. The spirit of this library is to offer easy-to-use helpers. If you need to optimize something: build it yourself. This is the Pareto Principle.
- `18.` Good. I have never been satisfied with the `lo.Entry` type. I should have used a `Tuple2`.
- `22.` Good. This is consistent with the rest of the library.

---

I wonder if it could be useful to duplicate some helpers to receive/return `iter.Seq2` iterators. Chaining multiple `lo` helpers without having the `iter.Seq2` variant seems complicated. It can be done in a different PR, but i'm asking this now to prevent any breaking change in method naming.

If we do such a variant, we have to think about the meaning of the 

Anyway, 2 more helpers would be nice: `Seq1ToSeq2` and `Seq2ToSeq1`, with `K` being a forever-increasing `int` in `Seq1ToSeq2`. WDYT?

Other proposal: add a `Void` or `Flush` method for consuming the iterator.

---

I just figured out that `iter` might not be the right name for this package, as suggested in #439. In fact, you cannot import twice a package called "iter" in a go file (`iter` std package + `github.com/samber/lo/iter`).

May I suggest "it"? Any idea?

---

No need to create examples on https://play.golang.org/. First, because we need to merge+release this contribution before, but also because I made an MCP server for creating all of them in 1 prompt.

I think it's time for me to create a dedicated website for documentation. Mixing basic helpers with mutable, parallel, and iterable helpers will make this README.md unreadable.

---

FYI, I will be releasing a reactive programming lib very soon (next week?), with an API similar to rxjs.

The API will be very different and stream-oriented (eg: `Sum` won't return a single value at the end, but as many values sent to it).

.github/workflows/lint.yml
```diff
@@ -11,12 +11,13 @@ jobs:
     steps:
       - uses: actions/setup-go@v5
         with:
-          go-version: 1.18
+          go-version: 1.23
```
samber: I would revert this change and add `//go:build go1.23` directive on top of each file

.github/workflows/lint.yml
```diff
@@ -11,12 +11,13 @@ jobs:
...
           stable: false
       - uses: actions/checkout@v5
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
+          version: 'v2.4'
```
samber: Why not latest ?

it/channel.go
```diff
@@ -0,0 +1,29 @@
+package iter
+
+import "iter"
+
+// ToChannel returns a read-only channel of collection elements.
+func ToChannel[T any](bufferSize int, collection iter.Seq[T]) <-chan T {
```
samber: As discussed in the PR, such helper may have a variant `ToChannel2` receiving `iter.Seq2`.

iter/type_manipulation.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+// FromAny returns a sequence with all elements mapped to a type.
+// Panics on type conversion failure.
+func FromAny[T any](collection iter.Seq[any]) iter.Seq[T] {
+	return func(yield func(T) bool) {
+		for item := range collection {
+			if t, ok := item.(T); !ok {
+				panic("iter.FromAny: type conversion failed")
```
samber: Todo: if we change the name of this sub-package, don't forget to update this.

Also: it should not be confused with `iter` package of the Go std lib.

it/seq.go
```diff
@@ -0,0 +1,813 @@
...
+		if ok {
+			result[k] = v
+		}
+	}
+
+	return result
+}
+
+// Keyify returns a map with each unique element of the sequence as a key.
+func Keyify[T comparable](collection iter.Seq[T]) map[T]struct{} {
```
samber, line 455: I don't know if Keyify should collect and return a map or return a `iter.Seq2[T, struct{}]` instead.

WDYT ?

Also, I think we can add an "Invert" helper: `iter.Seq2[K, V] -> iter.Seq2[V, K]`, similar to the `lo.Invert` helper manipulating maps.

it/map.go
```diff
@@ -0,0 +1,201 @@
...
+	out := make(Map)
+
+	for item := range maps {
+		for k := range item {
+			out[k] = item[k]
+		}
+	}
+
+	return out
+}
```
samber:
```suggestion
func Assign[K comparable, V any, Map ~map[K]V](maps ...iter.Seq[Map]) Map {
	out := make(Map)

	for iterator := range maps {
	    for item := range iterator {
		   for k := range item {
			   out[k] = item[k]
		   }
		}
	}

	return out
}
```

it/map.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+		if len(result) > 0 {
+			yield(result)
+		}
+	}
+}
+
+// FromMap transforms a map into a sequence based on specified iteratee.
+func FromMap[K comparable, V, R any](in map[K]V, iteratee func(key K, value V) R) iter.Seq[R] {
```
samber: The name of this helper looks not very self-explaining. Does it convert a map into something or does it perform a "map" operation as mentioned in the comment ?

I would name it MapToSeq. We already have MapToSlice in the base package. This helper would be the opposite of iter.Associate.

it/map.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+	}
+}
+
+// FilterFromMap transforms a map into a sequence based on specified iteratee.
+// The iteratee returns a value and a boolean. If the boolean is true, the value is added to the result sequence.
+// If the boolean is false, the value is not added to the result sequence.
+// The order of the keys in the input map is not specified and the order of the keys in the output sequence is not guaranteed.
+func FilterFromMap[K comparable, V, R any](in map[K]V, iteratee func(key K, value V) (R, bool)) iter.Seq[R] {
```
samber: Rename to FilterMapToSeq ?

it/map.go
```diff
@@ -0,0 +1,201 @@
...
+		for k := range in {
+			if v, ok := iteratee(k, in[k]); ok && !yield(v) {
+				return
+			}
+		}
+	}
+}
+
+// FilterKeys transforms a map into a sequence based on predicate returns true for specific elements.
+// It is a mix of lo.Filter() and lo.Keys().
```
samber:
```suggestion
// It is a mix of iter.Filter() and iter.Keys().
```

it/map.go
```diff
@@ -0,0 +1,201 @@
...
+		for k := range in {
+			if predicate(k, in[k]) && !yield(k) {
+				return
+			}
+		}
+	}
+}
+
+// FilterValues transforms a map into a sequence based on predicate returns true for specific elements.
+// It is a mix of lo.Filter() and lo.Values().
```
samber:
```suggestion
// It is a mix of iter.Filter() and iter.Values().
```

it/seq.go
```diff
@@ -0,0 +1,813 @@
...
+
+	for item := range collection {
+		result[item]++
+	}
+
+	return result
+}
+
+// CountValuesBy counts the number of each element returned from mapper function.
+// Is equivalent to chaining lo.Map and lo.CountValues.
```
samber:
```suggestion
// Is equivalent to chaining iter.Map and iter.CountValues.
```

.github/workflows/lint.yml
NathanBaulch: See #671 - I wanted all my PRs to pass the lint test but changes to golangci-lint released a few days ago caused new failures. It looks like you merge PRs even if the CI actions aren't passing?

.github/workflows/lint.yml
NathanBaulch: Reverted.

.github/workflows/lint.yml
NathanBaulch: Removed.

iter/type_manipulation.go
NathanBaulch: Done.

it/channel.go
NathanBaulch: What would this return? `<-chan lo.Entry[K, V]` or `<-chan lo.Tuple2[K, V]` maybe?

it/seq.go
NathanBaulch: Regarding `Keyify`, in my experience 99% of the time it's specifically used to create a map of keys for quick lookups, for example in the new `TrimLeft` and `TrimRight` helpers. There's also no perf benefit in switching to `iter.Seq2` because internally it still needs to scan the full sequence into a map to remove duplicates.
If a user wants an `iter.Seq2[T, struct{}]` then they can chain it without any perf cost using `Entries(Keyify(c))`.

iter/type_manipulation.go
NathanBaulch: I was originally thinking of aliasing the two stdlib iter types so users only need to import our package, however I don't believe generic type aliases are supported in Go 1.23.
```golang
package iter
import "iter"
type Seq[V any] = iter.Seq[V]
type Seq2[K, V any] = iter.Seq2[K, V]
```

it/map.go
NathanBaulch: Done.

it/map.go
NathanBaulch: Good catch, fixed.

it/seq.go
NathanBaulch: Fixed.

it/map.go
NathanBaulch: Fixed.

it/seq.go
NathanBaulch: Regarding `Invert`, done!

it/map.go
NathanBaulch: I removed the word "Slice" from all functions (except the actual `Slice` helper) because I assumed it would be implied by the "iter" package prefix. Here's the full list:
1. `lo.ChannelToSlice` -> `iter.FromChannel`
2. `lo.CoalesceSlice` -> `iter.Coalesce`
3. `lo.CoalesceSliceOrEmpty` -> `iter.CoalesceOrEmpty`
4. `lo.FilterMapToSlice` -> `iter.FilterFromMap`
5. `lo.FilterSliceToMap` -> `iter.FilterToMap`
6. `lo.FromAnySlice` -> `iter.FromAny`
7. `lo.FromSlicePtr` -> `iter.FromPtr`
8. `lo.FromSlicePtrOr` -> `iter.FromPtrOr`
9. `lo.MapToSlice` -> `iter.FromMap`
10. `lo.SliceToChannel` -> `iter.ToChannel`
11. `lo.SliceToMap` -> `iter.ToMap`
12. `lo.ToAnySlice` -> `iter.ToAny`
13. `lo.ToSlicePtr` -> `iter.ToPtr`

Do you only want `FromMap` renamed with the "Seq" convention, some others too, or maybe all of them?

it/map.go
NathanBaulch: See my followup question above.

NathanBaulch:
> Since you added `GroupBy`+`PartitionBy`+`Mode`, i don't feel uncomfortable adding such helpers, especially if this is properly documented.

Just to clarify, you're saying I **should** implement `Reverse`, `ReduceRight`, `Sample*`, `Shuffle` (with a disclaimers in the docs) even though they need to scan the full input sequence?

> If we do such a variant, we have to think about the meaning of the

Heh, where was this sentence going?

> 2 more helpers would be nice: `Seq1ToSeq2` and `Seq2ToSeq1`, with `K` being a forever-increasing `int` in `Seq1ToSeq2`.

Would we need two versions of `Seq2ToSeq1`, one for keys and one for values?

> Other proposal: add a `Void` or `Flush` method for consuming the iterator.

I would propose `Drain`.

samber:
> Just to clarify, you're saying I should implement Reverse, ReduceRight, Sample*, Shuffle (with a disclaimers in the docs) even though they need to scan the full input sequence?

Yes. Some helpers you built scan the full sequence as well. I would consider that developers understand that reversing a lazily produced slice is not possible without reading it 100%.

> Heh, where was this sentence going?

I don't remember 😅

> Would we need two versions of Seq2ToSeq1, one for keys and one for values?
 
Oh yes, definitely.

> I would propose Drain.

Perfect!

it/channel.go
samber: `<-chan lo.Tuple2[K, V]` seems future-proof if we remove `lo.Entry` in v2

it/seq.go
samber: Ok 👌

NathanBaulch:
> Some helpers you built scan the full sequence as well.

I just realized that I haven't explained how I decided which full-scan helpers to exclude. It's basically about memory - I skipped the ones that need to scan the full sequence into a slice internally. So `Sum`, `Reduce`, `IsSorted` are fine, but something like `Reverse` is just going to be a thin wrapper around the slice based `Reverse` helper.

```golang
func Reverse[T any, I ~func(func(T) bool)](collection I) I {
	s := slices.Collect(iter.Seq[T](collection))
	mutable.Reverse(s)
	return I(slices.Values(s))
}
```

Anyway, happy to add them with a suitable disclaimer in the docs.

it/channel.go
NathanBaulch: Done.

it/map.go
NathanBaulch: I've renamed these to use the "Seq" convention. Can roll back if needed.

samber:
We have too many flaky tests in this project... 🤬

image

samber:
Yes, accumulating items before returning the result is definitely a memory issue. But it would have been a problem for GroupBy and PartitionBy as well.

Let's add these helpers, with proper comments. I'm going to check if we have more methods in that situation in my next review.

it/intersect.go
```diff
@@ -0,0 +1,216 @@
...
+		if predicate(item) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// Intersect returns the intersection between two collections.
+func Intersect[T comparable, I ~func(func(T) bool)](list1, list2 I) I {
```
samber: add a comment: might cause a memory leak on long sequences

it/intersect.go
```diff
@@ -0,0 +1,216 @@
...
+		for item := range list2 {
+			if _, ok := seen[item]; ok && !yield(item) {
+				return
+			}
+		}
+	}
+}
+
+// Union returns all distinct elements from given collections.
+func Union[T comparable, I ~func(func(T) bool)](lists ...I) I {
```
samber, line 136: add a comment: might cause a memory leak on long sequences

it/intersect.go
```diff
@@ -0,0 +1,216 @@
...
+// ElementsMatch returns true if lists contain the same set of elements (including empty set).
+// If there are duplicate elements, the number of occurrences in each list should match.
+// The order of elements is not checked.
+func ElementsMatch[T comparable](list1, list2 iter.Seq[T]) bool {
+	return ElementsMatchBy(list1, list2, func(item T) T { return item })
+}
+
+// ElementsMatchBy returns true if lists contain the same set of elements' keys (including empty set).
+// If there are duplicate keys, the number of occurrences in each list should match.
+// The order of elements is not checked.
```
samber, line 185: add a comment: might cause a memory leak on long sequences

it/intersect.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+			i++
+		}
+	}
+}
+
+// ElementsMatch returns true if lists contain the same set of elements (including empty set).
+// If there are duplicate elements, the number of occurrences in each list should match.
+// The order of elements is not checked.
+func ElementsMatch[T comparable](list1, list2 iter.Seq[T]) bool {
```
samber, line 179: add a comment: might cause a memory leak on long sequences

it/map.go
```diff
@@ -0,0 +1,216 @@
...
+			if !yield(v, k) {
+				return
+			}
+		}
+	}
+}
+
+// Assign merges multiple sequences of maps from left to right.
+func Assign[K comparable, V any, Map ~map[K]V](maps ...iter.Seq[Map]) Map {
+	out := make(Map)
```
samber, line 123: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+			}
+			i++
+		}
+	}
+}
+
+// UniqMap manipulates a sequence and transforms it to a sequence of another type with unique values.
+func UniqMap[T any, R comparable](collection iter.Seq[T], iteratee func(item T) R) iter.Seq[R] {
+	return func(yield func(R) bool) {
+		seen := make(map[R]struct{})
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+				seen[r] = struct{}{}
+			}
+		}
+	}
+}
+
+// UniqMapI manipulates a sequence and transforms it to a sequence of another type with unique values.
+func UniqMapI[T any, R comparable](collection iter.Seq[T], iteratee func(item T, index int) R) iter.Seq[R] {
+	return func(yield func(R) bool) {
+		seen := make(map[R]struct{})
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+			}
+		}
+	}
+}
+
+// Uniq returns a duplicate-free version of a sequence, in which only the first occurrence of each element is kept.
+// The order of result values is determined by the order they occur in the sequence.
+func Uniq[T comparable, I ~func(func(T) bool)](collection I) I {
+	return func(yield func(T) bool) {
+		seen := make(map[T]struct{})
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		}
+	}
+}
+
+// UniqBy returns a duplicate-free version of a sequence, in which only the first occurrence of each element is kept.
+// The order of result values is determined by the order they occur in the sequence. It accepts `iteratee` which is
+// invoked for each element in the sequence to generate the criterion by which uniqueness is computed.
+func UniqBy[T any, U comparable, I ~func(func(T) bool)](collection I, iteratee func(item T) U) I {
+	return func(yield func(T) bool) {
+		seen := make(map[U]struct{})
```
samber, line 227: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+			if !yield(item) {
+				return
+			}
+		}
+	}
+}
+
+// GroupBy returns an object composed of keys generated from the results of running each element of collection through iteratee.
+func GroupBy[T any, U comparable](collection iter.Seq[T], iteratee func(item T) U) map[U][]T {
+	result := map[U][]T{}
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+	for item := range collection {
+		key := iteratee(item)
+
+		result[key] = append(result[key], item)
+	}
+
+	return result
+}
+
+// GroupByMap returns an object composed of keys generated from the results of running each element of collection through iteratee.
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		if len(newSlice) > 0 {
+			yield(newSlice)
+		}
+	}
+}
+
+// PartitionBy returns a sequence of elements split into groups. The order of grouped values is
+// determined by the order they occur in collection. The grouping is generated from the results
+// of running each element of collection through iteratee.
+func PartitionBy[T any, K comparable](collection iter.Seq[T], iteratee func(item T) K) [][]T {
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+					done++
+				} else if !yield(t) {
+					return
+				}
+			}
+		}
+	}
+}
+
+// Shuffle returns a sequence of shuffled values. Uses the Fisher-Yates shuffle algorithm.
```
samber, line 354: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+func Shuffle[T any, I ~func(func(T) bool)](collection I) I {
+	slice := slices.Collect((iter.Seq[T])(collection))
+	mutable.Shuffle(slice)
+	return I(slices.Values(slice))
+}
+
+// Reverse reverses a sequence so that the first element becomes the last, the second element becomes the second to last, and so on.
```
samber, line 363: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+	return func(yield func(T) bool) {
+		for i := range count {
+			if !yield(predicate(i)) {
+				return
+			}
+		}
+	}
+}
+
+// KeyBy transforms a sequence to a map based on a pivot callback.
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		result[k] = item
+	}
+
+	return result
+}
+
+// Associate returns a map containing key-value pairs provided by transform function applied to elements of the given sequence.
+// If any of two pairs have the same key the last one gets added to the map.
+// The order of keys in returned map is not specified and is not guaranteed to be the same from the original sequence.
+func Associate[T any, K comparable, V any](collection iter.Seq[T], transform func(item T) (K, V)) map[K]V {
```
samber, line 416: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		result[k] = v
+	}
+
+	return result
+}
+
+// SeqToMap returns a map containing key-value pairs provided by transform function applied to elements of the given sequence.
+// If any of two pairs have the same key the last one gets added to the map.
+// The order of keys in returned map is not specified and is not guaranteed to be the same from the original sequence.
+// Alias of Associate().
```
samber, line 430: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		k, v, ok := transform(item)
+		if ok {
+			result[k] = v
+		}
+	}
+
+	return result
+}
+
+// Keyify returns a map with each unique element of the sequence as a key.
```
samber, line 453: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+// Alias of Associate().
+func SeqToMap[T any, K comparable, V any](collection iter.Seq[T], transform func(item T) (K, V)) map[K]V {
+	return Associate(collection, transform)
+}
+
+// FilterSeqSeqToMap returns a map containing key-value pairs provided by transform function applied to elements of the given sequence.
+// If any of two pairs have the same key the last one gets added to the map.
+// The order of keys in returned map is not specified and is not guaranteed to be the same from the original sequence.
+// The third return value of the transform function is a boolean that indicates whether the key-value pair should be included in the map.
+func FilterSeqSeqToMap[T any, K comparable, V any](collection iter.Seq[T], transform func(item T) (K, V, bool)) map[K]V {
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+				dropping = false
+			}
+			if !dropping && !yield(item) {
+				return
+			}
+		}
+	}
+}
+
+// DropRightWhile drops elements from the end of a sequence while the predicate returns true.
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		for item := range collection {
+			if i >= n && !yield(item) {
+				return
+			}
+			i++
+		}
+	}
+}
+
+// DropRight drops n elements from the end of a sequence.
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+			count++
+		}
+	}
+
+	return count
+}
+
+// CountValues counts the number of each element in the collection.
+func CountValues[T comparable](collection iter.Seq[T]) map[T]int {
+	result := make(map[T]int)
```
samber: add a comment: might cause a memory leak on long sequences

it/seq.go
```diff
@@ -0,0 +1,849 @@
...
+		result[item]++
+	}
+
+	return result
+}
+
+// CountValuesBy counts the number of each element returned from mapper function.
+// Is equivalent to chaining Map and CountValues.
+func CountValuesBy[T any, U comparable](collection iter.Seq[T], mapper func(item T) U) map[U]int {
+	result := make(map[U]int)
```
samber, line 624: add a comment: might cause a memory leak on long sequences

samber:
What do you think about renaming:
- DropRight by DropLast
- ReduceRight by ReduceLast

?

"Left" and "right" are applicable to slices, but not iterators.

In the reactive programming vocabulary, we don't use left/right.

NathanBaulch:
Pretty sure I've addressed all outstanding feedback - time for an update!

1. Added `Seq` to the names of helpers corresponding to existing helpers with `Slice` in their name.
2. Replaced `Right` with `Last` and `Left` with `First`.
3. Implemented new helpers `Drain`, `Seq2Channel`, `SeqToSeq2`, `Seq2KeyToSeq`, `Seq2ValueToSeq`.
4. Ported helpers that iterate the full sequence `Reverse`, `ReduceRight`, `Sample`, `Shuffle`.
5. Reviewed all helpers and added documentation if they fully iterate before returning or they allocate internal slices/maps.
6. Ported recent additions `HasPrefix`, `HasSuffix`, `Cut*`, `Trim*`.
   `Cut` isn't supported since it would need to return multiple sequences, similar to `Difference`.
7. Implemented new helper `Length` which corresponds to `len(myslice)`. Can rename to `Len` if preferred.
8. `Intersect` is now variadic, taking any number of input sequences. Perhaps the slice version can also in v2.
9. Renamed `IsSortedByKey` to `IsSortedBy` to be more consistent with other `*By` helpers. Another idea for v2.
10. Renamed all map/filter callback parameters to transform/predicate respectively. "iteratee" is a pretty obscure word and hasn't been used consistently. For example, `lo.ForEachWhile` and `lo.RepeatBy` callbacks are named incorrectly.
11. Significantly reduced duplicate code by using helpers inside helpers. Iterators are well suited to this since they don't allocate return slices.
    For example, `Subset` calls `Slice`, `Replace`, `ToAnySeq` call `Map`, `Compact` calls `Filter`, etc.
12. Lots of performance tweaks, especially around reducing allocations and multiple iteration.
13. `DropByIndex` and `WithoutNth` are identical, `Contains` is the same as `Some` if only passed a single item.

.github/workflows/lint.yml
```diff
@@ -17,6 +17,7 @@ jobs:
...
```
samber:
```suggestion
```

samber:
Fantastic!

It's time to merge this PR. Thanks for your work on this long-awaited feature!



Next steps for me:
- release of v1.52.0
- publish lo.samber.dev
- fix these flaky tests

it/seq.go
mabrarov: Considering https://github.com/samber/lo/pull/697, don't we need to add into this file:

1. AssociateI
2. SeqToMapI
3. FilterSeqToMap

?

mabrarov: If it will be found reasonable, then here is PR: https://github.com/samber/lo/pull/698

it/seq.go
```diff
@@ -0,0 +1,904 @@
...
+	}
+
+	return result
+}
+
+// Associate returns a map containing key-value pairs provided by transform function applied to elements of the given sequence.
+// If any of two pairs have the same key the last one gets added to the map.
+// The order of keys in returned map is not specified and is not guaranteed to be the same from the original sequence.
+// Will iterate through the entire sequence.
+func Associate[T any, K comparable, V any](collection iter.Seq[T], transform func(item T) (K, V)) map[K]V {
```
mabrarov, line 416: Considering https://github.com/samber/lo/pull/697, don't we need to add into this file:

1. AssociateI
2. SeqToMapI
3. FilterSeqToMap

?

mabrarov: If it will be found reasonable, then here is PR: https://github.com/samber/lo/pull/698

samber:
Documentation is available here: https://lo.samber.dev/docs/iter/map

The runnable examples will be published on Go playground tomorrow or on Wednesday.

PR #678, opened 2025-09-24
fix: more consistent panic strings

NathanBaulch:
I noticed some inconsistencies in panic strings during my work on #672. I went with the pattern used in `lo.RandomString` and `o.ChunkString` with the function name prefix and lower case message. I don't mind what convention is used as long as it's consistent.
Note that I haven't touched the panics returned from `Must` which use a "must: " prefix in lowercase, since consumers might have recover logic that inspects this relatively common panic message.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
-          version: 'v2.4'
           args: --timeout 120s --max-same-issues 50
 
       - name: Bearer


.github/workflows/lint.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
+          version: 'v2.4'
```
samber: why not using latest version ?

samber (commented):
Hey @NathanBaulch 

Thanks for this fix. Just a little comment ;)

.github/workflows/lint.yml
samber: I removed this line to allow the merge. Feel free to reply to me in #672.

NathanBaulch: See #671 for an explanation.

PR #682, opened 2025-09-24
lint: fix CI warnings and avoid named return parameters

NathanBaulch:
The latest golangci-lint version now warns against naked returns. This PR fixes those and also removes unnecessary named return parameters in a few places. IMO named parameters should only be used on functions that use `recover()` and functions that have > 2 non-obvious return values.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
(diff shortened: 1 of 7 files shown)


NathanBaulch:
This also restores golangci-lint to latest version, as discussed in #671.

NathanBaulch:
#666 introduced quite a few linting issues. Shouldn't all PR checks be passing before a PR is merged?

samber:
Thanks for your changes.

PR #687, opened 2025-09-25
fix: switch 2 parameters from uint to int for consistency

NathanBaulch:
Per item 22 on #672, the `length` parameter on `Subset` and `Substring` has been changed to `int`.
In theory a breaking API change, but seems pretty unlikely anyone would be passing a length value >= 2^31.


samber:
I'm delaying the merge to v2.

PR #689, opened 2025-09-26
perf: avoid Keyify twice in Trim

NathanBaulch:
Chaining `TrimLeft` and `TrimRight` means the input collection will be unnecessarily "Keyified" twice.

Speaking of trim helpers, I was surprised to see that `TrimPrefix` trims all instances of the leading slice, not just the first. This is different from the stdlib `strings.TrimPrefix` and should probably be documented. Same with the suffix version.


NathanBaulch:
Actually, thinking about this more... Chaining `DropWhile` and `DropRightWhile` means allocating two slices which isn't ideal. This is now fixed.
Any helper that chains other helpers internally should take these kinds of costs into consideration.

PR #690, opened 2025-09-26
lint: enable gofumpt extra rules

NathanBaulch:
I'm personally a fan of the extra gofumpt rules. Feel free to close this PR if you think they're overly strict.


samber:
I prefer explicit parameters, but the Go community has decided to group parameters ;)

PR #691, opened 2025-09-26
lint: use Keyify internally

NathanBaulch:
There are 4 intersect helpers that could benefit from using the existing `Keyify` helper internally.

This is a question I have quite often whilst working on iterators support - when is it appropriate to use another helper and when is it better to implement from scratch?
There are lots of places where simple helpers like `Empty` or `Map` or even `Ternary` could be used. What about related helpers like like `Count` using `CountBy` internally, etc?


samber:
I'm pretty sure there is no difference in final asm code.

Go compiler inlines such methods when no side effect is detected.

PR #694, opened 2025-09-30
fix: omit duplicates in second Intersect list

NathanBaulch:
Duplicates passed to `Intersect` in `list1` are omitted and retained in `list2`. I believe this should be a symmetric function, producing identical results when the two input lists are swapped.


d-enk:
Fixed by https://github.com/samber/lo/pull/739

may be closed

Issue #601, opened 2025-02-26
Converting slice to map (new helper)


renom:
Input:
```go
[]string{"string 1", "string 2", "string 3"}
```

Output:
```go
map[int]string{0: "string 1", 1: "string 2", 2: "string 3"}
```

samber:
It already exists -> lo.SliceToMap ;)

mabrarov:
Hi @samber,

Unfortunately, [SliceToMap](https://github.com/samber/lo/blob/v1.51.0/slice.go#L399) (actually [Associate](https://github.com/samber/lo/blob/v1.51.0/slice.go#L383) which implements `SliceToMap`), as well as [FilterSliceToMap](https://github.com/samber/lo/blob/v1.51.0/slice.go#L407), doesn't pass index of original collection item to `transform` func:

```go
func Associate[T any, K comparable, V any](collection []T, transform func(item T) (K, V)) map[K]V {
	result := make(map[K]V, len(collection))

	for i := range collection {
		k, v := transform(collection[i])
		result[k] = v
	}

	return result
}
```

It makes `SliceToMap` insufficient (ineffective) in some cases, like the one which @renom described and which can be implemented only with additional temporary collection (https://go.dev/play/p/QH80tW3tM5K):

```go
list := []string{"string 1", "string 2", "string 3"}
result := lo.FromEntries(lo.Map(list, func(str string, index int) lo.Entry[int, string] {
	return lo.Entry[int, string]{
		Key:   index,
		Value: str,
	}
}))
```

IMHO, for all functions which take slice as an input (source) collection, we need to pass item index into transformation function (but it's too late to make such backward incompatible change, so we may need to consider introduction of new functions for slice → map conversions). In case of `SliceToMap` the new function could look like (https://go.dev/play/p/n1ehTa4lyjN):

```go
func SliceWithIndexToMap[T any, K comparable, V any](collection []T, transform func(item T, index int) (K, V)) map[K]V {
	result := make(map[K]V, len(collection))

	for index, item := range collection {
		k, v := transform(item, index)
		result[k] = v
	}

	return result
}
```

Thank you.

samber:
Thanks for your suggestion @mabrarov. Please open a new issue or at least reopen this one.

For such an helper, I would name it `lo.SliceToMapI`. This is the new default naming for "with index" variants.

mabrarov:
Hi @samber,

[...]

PR #697, opened 2025-10-02
feat: from slice to map with index passed

mabrarov:
Conversion from slice to map with index passed to transformation function (https://github.com/samber/lo/issues/601).

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -397,6 +397,21 @@ func Associate[T any, K comparable, V any](collection []T, transform func(item T
 	return result
 }
 
+// AssociateI returns a map containing key-value pairs provided by transform function applied to elements of the given slice.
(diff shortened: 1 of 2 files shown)


samber (approved):

slice.go
mabrarov, line 1: Hi @samber,

Considering implementation of similar functions in it/seq.go file of https://github.com/samber/lo/pull/672  (e.g. `it.Filter` and `it.FilterI`), do you find it reasonable to do the same here, i.e. implement `Associate` with `AssociateI`, `SliceToMap` with `SliceToMapI` and `FilterSliceToMap` with `FilterSliceToMapI`?

Thank you.

mabrarov: If it will be found reasonable, then here is PR: https://github.com/samber/lo/pull/698

PR #711, opened 2025-10-12
fix(it.Mode): align behavior with lo.Mode and ensure consistent slice…

### LINKED ISSUE #709: Different behavior between `lo.Mode` / `it.Mode`
v1.52.0

```go
package main

import (
	"fmt"
	"slices"

	"github.com/samber/lo"
	"github.com/samber/lo/it"
)

func main() {
	arr := []int{1, 1, 2, 2, 3, 3, 3}

	mode := lo.Mode(arr)
	fmt.Println(cap(mode), mode)

	mode = it.Mode(slices.Values(arr))
	fmt.Println(cap(mode), mode)
}
```

```
1 [3]
4 [3]
```

```go
mode = append(mode[:0], item)
``` 

is better (no new allocations) than

```go
mode = []T{item}
```

but need something like

```go
return mode[:len(mode):len(mode)]
```

or 

```go
return append(T{},mode[:len(mode)]) // discard tail
```

intojhanurag:
fix #709 
Fixed inconsistent slice capacity between lo.Mode and it.Mode.
Reused slice memory with append(mode[:0], item) and added a test
to ensure len(result) == cap(result).

image

### DIFF
--- a/math_test.go
+++ b/math_test.go
@@ -191,3 +191,15 @@ func TestMode(t *testing.T) {
 	is.Empty(result4)
 	is.Equal([]int{1, 2, 3, 4, 5, 6, 7, 8, 9}, result5)
 }
+
+func TestModeCapacityConsistency(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
+	arr := []int{1, 1, 2, 2, 3, 3, 3}
+
+	result := Mode(arr)
+
+	is.Equal([]int{3}, result, "Mode should return correct mode value")
+	is.Equal(len(result), cap(result), "Mode slice capacity should match its length")
+}


it/math_test.go
```diff
@@ -177,3 +177,15 @@ func TestMode(t *testing.T) {
 	is.Empty(result4)
 	is.Equal([]int{1, 2, 3, 4, 5, 6, 7, 8, 9}, result5)
 }
+
+func TestModeCapacityConsistency(t *testing.T) {
```
samber, line 181: can you replicate the test in core package please ?

intojhanurag:
Hey @samber , This is ready for review.

it/math_test.go
intojhanurag: done!

PR #713, opened 2025-10-13
Fix references to "lo" instead of "ro"

pgr0ss:
I noticed the docs site referenced `samber/ro` instead of `samber/lo` in a few places, so I updated the links. However, I wasn't sure about the `llms.txt` change. Is there a different project we should be linked to?


samber:
Oooops, thanks for the report !

I reverted your change on llms.txt, since the project will be published tomorrow. ;)

I created both documentation websites simultaneously (lo+ro).

PR #714, opened 2025-10-15
Add a Concat slice function.

FGasper:
This is a variadic wrapper around Flatten that allows callers to assemble one slice from many slices without first joining them into a slice-of-slices.

### DIFF
--- a/slice_test.go
+++ b/slice_test.go
@@ -363,7 +363,7 @@ func TestConcat(t *testing.T) {
 
 	result1 := Concat([][]int{{0, 1}, {2, 3, 4, 5}}...)
 
-	is.Equal(result1, []int{0, 1, 2, 3, 4, 5})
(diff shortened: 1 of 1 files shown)


slice_test.go
```diff
@@ -357,6 +357,20 @@ func TestFlatten(t *testing.T) {
 	is.IsType(nonempty, allStrings, "type preserved")
 }
 
+func TestConcat(t *testing.T) {
+	t.Parallel()
+	is := assert.New(t)
+
+	result1 := Concat([][]int{{0, 1}, {2, 3, 4, 5}}...)
+
+	is.Equal(result1, []int{0, 1, 2, 3, 4, 5})
```
samber:
```suggestion
	is.Equal([]int{0, 1, 2, 3, 4, 5}, result1)
```

PR #724, opened 2025-11-05
fix: update Play links for Async functions in concurrency.go

Weeranieb:


samber:
Thanks!

Im going to make the change in the doc.

PR #729, opened 2025-11-06
lint: use consistent short slice creation syntax

NathanBaulch:
There are a few different patterns used to create slices. This PR makes them all consistent.
This is something I noticed whilst building the `it` subpackage.


samber:
There is actually a small difference between `var a []A` and `a := []A{}`. When you compare `a` to `nil`, the first case will output `true`.

https://go.dev/play/p/ZT1csskjjK_b

NathanBaulch:
I'm aware of that but have yet to find a real-world situation (other than edge cases using reflection) where this was an issue.
As usual, I advocate for consistency. If we want to avoid nil slices then we should do it everywhere and validate with unit tests.

PR #730, opened 2025-11-06
lint: fix inconsistent callback function parameter names

NathanBaulch:
Quite a few callback function parameters are incorrectly named. This PR attempts to fix a few of these. This is something I noticed whilst building the `it` subpackage.

There are always exceptions but...
* if a callback returns a single bool then it should probably be called "predicate".
* if a callback is used to change a collection element into something else then it should probably be called "transform".
* if a callback returns nothing (void) then it should probably be called "callback".
* Bonus: replace "cb" with "callback" for consistency.

Also, I've never seen the word "iteratee" used anywhere else before (as a native English speaker). I would propose renaming them all to something more meaningful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,9 +140,10 @@ Supported helpers for slices:
 - [Slice](#slice)
 - [Replace](#replace)
 - [ReplaceAll](#replaceall)
+- [Clone](#clone)
 - [Compact](#compact)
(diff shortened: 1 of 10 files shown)


samber:
I also added: https://github.com/samber/lo/pull/738

PR #737, opened 2025-11-06
refactor: reuse helpers where possible

NathanBaulch:
This is a continuation of #691, which in turn was done in response to the huge number of code reuse opportunities I found whilst building the `it` subpackage.

This PR manages to reduce the total net SLOC by 409 lines (+103 -512) by carefully using existing helpers wherever it can be done efficiently without additional memory allocations or unnecessary compute.

I've utilized first-class functions quite a lot, especially one of my favorite patterns involving `Partial`, `HasKey` and `Keyify` to build a compact and efficient predicate, eg:

```go
DropWhile(collection, Partial(HasKey, Keyify(cutset)))
```

I've also added internal `_map`, `_filter` and `_reject` functions that omit the index argument on callbacks to further improve first-class function syntax opportunities. Let me know if these should be renamed.

### DIFF
--- a/intersect.go
+++ b/intersect.go
@@ -21,7 +21,7 @@ func ContainsBy[T any](collection []T, predicate func(item T) bool) bool {
 // Every returns true if all elements of a subset are contained in a collection or if the subset is empty.
 // Play: https://go.dev/play/p/W1EvyqY6t9j
 func Every[T comparable](collection, subset []T) bool {
-	return len(subset) == 0 || EveryBy(subset, Partial(HasKey, Keyify(collection)))
+	return len(subset) == 0 || EveryBy(subset, Partial(HasKey[T, struct{}], Keyify(collection)))
 }
 
 // EveryBy returns true if the predicate returns true for all elements in the collection or if the collection is empty.
@@ -40,7 +40,7 @@ func EveryBy[T any](collection []T, predicate func(item T) bool) bool {
 // If the subset is empty Some returns false.
 // Play: https://go.dev/play/p/Lj4ceFkeT9V
 func Some[T comparable](collection, subset []T) bool {
-	return len(subset) > 0 && SomeBy(collection, Partial(HasKey, Keyify(subset)))
+	return len(subset) > 0 && SomeBy(collection, Partial(HasKey[T, struct{}], Keyify(subset)))
 }
 
 // SomeBy returns true if the predicate returns true for any of the elements in the collection.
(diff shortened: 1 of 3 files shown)


find.go
```diff
@@ -263,49 +210,13 @@ func FindDuplicatesBy[T any, U comparable, Slice ~[]T](collection Slice, iterate
...
-		item := collection[i]
-
-		if item < mIn {
-			mIn = item
-			index = i
-		}
-	}
-
-	return mIn, index
+	return MinIndexBy(collection, func(a, b T) bool { return a < b })
```
samber, line 219: I just made a quick benchmark to check if inlining works well.

```
func MinIndex2[T constraints.Ordered](collection []T) (T, int) {
	return MinIndexBy(collection, func(a, b T) bool { return a < b })
}

func BenchmarkMinIndex(b *testing.B) {
	arr := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

	b.Run("lo.MinIndex", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			_, _ = MinIndex(arr)
		}
	})
	b.Run("lo.MinIndexNew", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			_, _ = MinIndex2(arr)
		}
	})
}
```

I did not investigate more, but the result is not very good:

```
goos: darwin
goarch: arm64
pkg: github.com/samber/lo
cpu: Apple M3
=== RUN   BenchmarkMinIndex
BenchmarkMinIndex
=== RUN   BenchmarkMinIndex/lo.MinIndex
BenchmarkMinIndex/lo.MinIndex
BenchmarkMinIndex/lo.MinIndex-8                 329917464                3.370 ns/op           0 B/op          0 allocs/op
=== RUN   BenchmarkMinIndex/lo.MinIndexNew
BenchmarkMinIndex/lo.MinIndexNew
BenchmarkMinIndex/lo.MinIndexNew-8              261727790                4.435 ns/op           0 B/op          0 allocs/op
PASS
```

NathanBaulch: Ah, that's disappointing. Hopefully I can look deeper into this sometime.

Side note, it would be nice if we had a comprehensive suite of benchmarks with realistic examples that could be automatically checked for perf regressions. Someday...

NathanBaulch:
It's much worse on my machine! :sweat: 
I added a third test with inlining explicitly disabled and it's twice as bad again. So the callback funcs _are_ inlined but clearly aren't being further optimized like in the original helper (bounds checks, branch prediction or some other compiler magic).

```
goos: linux
goarch: amd64
pkg: github.com/samber/lo
cpu: Intel(R) Core(TM) Ultra 9 275HX
BenchmarkMinIndex/lo.MinIndex-24              1000000000         0.775 ns/op
BenchmarkMinIndex/lo.MinIndexNew-24            317205955         3.914 ns/op
BenchmarkMinIndex/lo.MinIndexNoInlining-24     134372474         8.784 ns/op
```

samber:
Ok, so I'm closing this PR for now. Improving readability is cool, but for such simple helpers, I consider the performance gain as more important.

FYI, some helpers have been combined for performance reasons: Filter+Map -> FilterMap.

PR #739, opened 2025-11-06
Refactoring lo.IntersectBy + adding loit.IntersectBy + adding doc

samber:
@NathanBaulch FYI

### DIFF
--- a/lo_example_test.go
+++ b/lo_example_test.go
@@ -3343,7 +3343,15 @@ func ExampleCrossJoinBy9() {
 }
 
 func ExampleIntersect() {
-	fmt.Printf("%v", Intersect([]int{0, 3, 5, 7}, []int{3, 5}, []int{0, 1, 2, 0, 3, 0}))
+	result := Intersect([]int{0, 3, 5, 7}, []int{3, 5}, []int{0, 1, 2, 0, 3, 0})
+	fmt.Printf("%v", result)
 	// Output:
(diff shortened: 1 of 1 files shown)


NathanBaulch:
LGTM! My only complaint would be the duplicated code between `Intersect` and `IntersectBy` but we've got #737 to track that.

samber:
Yes, same feeling for me.

d-enk:
It looks much better

```go
seen := make(map[T]bool, len(lists[len(lists)-1]))
for _, item := range lists[len(lists)-1] {
	seen[item] = true
}

for i := len(lists) - 2; i > 0; i-- {
...
}

result := make(Slice, 0, len(seen))
...
return result
```

than

```go
for i := len(lists) - 1; i >= 0; i-- {
	if i == len(lists)-1 {
		for _, item := range lists[i] {
			seen[item] = true
		}
		continue
	}

	if i == 0 {
		result := make(Slice, 0, len(seen))
		...
		return result
	}
	
	...
}

return Slice{}
```

Or am I missing something?

d-enk:
@samber 

Or at all

```go
seen := make(map[T]bool, len(lists[0]))
for _, item := range lists[0] {
	seen[item] = true
}

for i := 1; i < len(lists) - 1; i++ {
...
}


for item := range lists[len(lists) - 1] {
...
}

return result
```

PR #752, opened 2025-12-10
Must support Custom error handler.

RelicOfTesla:
Support user custom error handle in Must() ( like attach the callstack information );
Support Must("foo", customErr{}) to restore customErr type(ErrorsAs).

// Deprecated old pull #419 , Adopting a less invasive code approach


samber:
Hey @RelicOfTesla !

Thanks for that suggestion. I Confirm this is much less invasive !

PR #755, opened 2025-12-15
Support Custom Assert

RelicOfTesla:
user custom Assert/Assertf


samber:
I like it!

I'm going to add it to the documentation.

PR #756, opened 2025-12-29
refactor: improve lo[it].Intersect[By] readability and performance

d-enk:
Preallocate seen map capacity.

Extract first and last list handling out of the main loop.

Remove redundant map traversal by combining the reset and delete passes into a single loop. Values are now initialized to false and toggled to true when found, allowing the cleanup pass to both reset (true→false) and delete in one iteration.

Move len(seen) == 0 check into loop condition, allowing early exit before processing next list when intersection becomes empty (skips redundant first iteration).

### DIFF
--- a/intersect.go
+++ b/intersect.go
@@ -121,10 +121,6 @@ func Intersect[T comparable, Slice ~[]T](lists ...Slice) Slice {
 		return Slice{}
 	}
 
-	if len(lists) == 1 {
-		return lists[0]
-	}
-
 	last := lists[len(lists)-1]
 
 	seen := make(map[T]bool, len(last))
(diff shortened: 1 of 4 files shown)


d-enk:
Based on 
- https://github.com/samber/lo/pull/739#issuecomment-3501557156
- https://github.com/samber/lo/pull/739#issuecomment-3501582676

d-enk:
```go
if len(lists) == 1 {
	return lists[0]
}
```

Also, it seems that this early return case is harmful.
- not a copy
- not unique elements

samber:
Thanks @d-enk for this new contrib.

Yes, your implem seems better and leverage CPU cache.

Can you make the last suggested fix ?

PR #757, opened 2025-12-30
fix(iter/tuples): support break iteration over Zip[By] seq

d-enk:
fix all https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/yield


samber:
Thank you for your careful attention !!

staticchecks or govet should handle that case.

PR #760, opened 2026-01-06
feat: add Take, TakeWhile, FilterTake, Window, and Sliding functions

juliazadorozhnaya:
Add five new slice manipulation functions with tests, examples, and documentation.

### Take
Safely takes the first n elements from a slice. Handles edge cases and creates a copy to avoid mutating the original.

**Production use cases:**
- Pagination: Get first page of results
- Top N items: Extract top 10 users, products, or metrics
- Batch processing: Process first 100 tasks from queue

### TakeWhile
Takes elements from the beginning while the predicate returns true. Useful for processing ordered data until a condition is met.

**Production use cases:**
- Process successful transactions until first failure
- Parse valid tokens until first invalid one

### FilterTake
Filters elements and takes the first n matches. More efficient than chaining Filter and Take, as it stops after finding n matches.

**Production use cases:**
- Search: Find first 5 active users (stops early, doesn't scan entire list)

### Window
Creates overlapping sliding windows of a given size (step=1). Each window overlaps with the previous one by size-1 elements.

**Production use cases:**
- Time series analysis: Calculate 7-day moving averages
- Anomaly detection: Compare adjacent time periods

### Sliding
Creates sliding windows with configurable step size. More flexible than Window - can create overlapping or non-overlapping windows.

**Production use cases:**
- Batch processing: Process data in chunks (step=size, like Chunk)
- Aggregation: Calculate averages over 5-minute windows with 1-minute steps
- Sampling: Extract every 10th element in batches

### DIFF
--- a/benchmark/slice_benchmark_test.go
+++ b/benchmark/slice_benchmark_test.go
@@ -207,19 +207,21 @@ func BenchmarkFilterTakeVsFilterAndTake(b *testing.B) {
 
 	b.Run("lo.TakeFilter", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
-			_ = lo.TakeFilter(ints, 5, func(v int, _ int) bool { return v%2 == 0 })
+			_ = lo.TakeFilter(ints, 5, func(v, _ int) bool {
+				return v%2 == 0
+			})
 		}
 	})
 
 	b.Run("lo.Filter+lo.Take", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
-			_ = lo.Take(lo.Filter(ints, func(v int, _ int) bool { return v%2 == 0 }), 5)
+			_ = lo.Take(lo.Filter(ints, func(v, _ int) bool { return v%2 == 0 }), 5)
 		}
 	})
 
 	b.Run("lo.Filter+native_slice", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
-			filtered := lo.Filter(ints, func(v int, _ int) bool { return v%2 == 0 })
+			filtered := lo.Filter(ints, func(v, _ int) bool { return v%2 == 0 })
 			takeN := 5
 			if takeN > len(filtered) {
 				_ = filtered
--- a/lo_example_test.go
+++ b/lo_example_test.go
@@ -2423,9 +2423,11 @@ func ExampleTakeWhile() {
 func ExampleTakeFilter() {
 	list := []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
 
-	result := TakeFilter(list, 3, func(val int, index int) bool {
-		return val%2 == 0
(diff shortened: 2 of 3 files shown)


slice.go
```diff
@@ -563,6 +631,33 @@ func DropByIndex[T any, Slice ~[]T](collection Slice, indexes ...int) Slice {
 	return result
 }
 
+// FilterTake filters elements and takes the first n elements that match the predicate.
+// This is more efficient than chaining Filter and Take, as it stops after finding n matches.
+func FilterTake[T any, Slice ~[]T](collection Slice, n int, predicate func(item T, index int) bool) Slice {
```
samber: I would rename it to `TakeFilter`, since every combinaisons of helpers have such naming in `lo`.

Equivalent to calling `Take(Filter(...))`

samber: This helper is currently missing in the docs/data/directory.

samber (commented):
Thanks for this first contribution.

Please add the description of the helper to the README. The documentation website is still young. I'm going to remove this legacy doc as soon as the SEO improves.

I suppose we can create similar helpers in subpackages and especially `it`, but we can do it in a follow-up PR.

slice.go
```diff
@@ -283,6 +283,43 @@ func Concat[T any, Slice ~[]T](collections ...Slice) Slice {
...
+
+// Sliding creates a slice of sliding windows of a given size with a given step.
+// If step is equal to size, windows don't overlap (similar to Chunk).
+// If step is less than size, windows overlap.
+func Sliding[T any, Slice ~[]T](collection Slice, size, step int) []Slice {
+	if size <= 0 {
+		panic("lo.Sliding: size must be greater than 0")
+	}
+
+	if step <= 0 {
```
samber, line 304: I wonder if it should be greater to 1 instead of 0 ?

samber (commented):
Except 1 comment, it seems good to me.

Do you plan to add more things on this PR ? Would you like to remove the draft tag ?

Edit: Please fix the linter error

samber:
Please note that some of your helpers are already implemented in https://github.com/samber/ro, with a streaming paradigm.

slice.go
juliazadorozhnaya: Hi @samber! Thanks for the review.

The check if `step <= 0` is correct. This matches the pattern used for size in Chunk, Window, and Sliding (all use <= 0 with "must be greater than 0"), keeping the codebase consistent. The logic is equivalent to `step < 1` or `!(step >= 1)`, but `step <= 0` aligns with the existing style. If you prefer a different formulation (like `step < 1`), I'm happy to change it.

juliazadorozhnaya: I'm planning to add the iterator package implementations in a follow-up PR. I'll finish the current work and remove the draft tag within the next hour.

samber (approved):

PR #770, opened 2026-01-12
perf: use `Ternary()` instead `If().Else()` in `Range`/`RangeFrom`

d-enk:
No allocs in it.Range, it.RangeFrom

For Benchmarks like:

```go
func BenchmarkRange(b *testing.B) {
	for i := 0; i < b.N; i++ {
		for range Range(0) {
		}
	}
}
```

```go
pkg: github.com/samber/lo

            │   old.txt    │               new.txt                │
            │    sec/op    │    sec/op     vs base                │
RangeFrom-4   6.9015n ± 1%   0.2802n ± 1%  -95.94% (p=0.000 n=10)
Range-4       6.8400n ± 1%   0.2803n ± 0%  -95.90% (p=0.000 n=10)

            │   old.txt    │               new.txt               │
            │     B/op     │    B/op     vs base                 │
RangeFrom-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10)
Range-4       0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10)

            │   old.txt    │               new.txt               │
            │  allocs/op   │ allocs/op   vs base                 │
RangeFrom-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10)
Range-4       0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10)

pkg: github.com/samber/lo/it
            │    old.txt     │                new.txt                │
            │     sec/op     │    sec/op      vs base                │
RangeFrom-4   85.8800n ±  2%   0.2802n ±  1%  -99.67% (p=0.000 n=10)
Range-4       67.8400n ± 13%   0.2807n ± 32%  -99.59% (p=0.000 n=10)

            │  old.txt   │                new.txt                 │
            │    B/op    │   B/op     vs base                     │
RangeFrom-4   72.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)
Range-4       48.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

            │  old.txt   │                 new.txt                 │
            │ allocs/op  │ allocs/op   vs base                     │
RangeFrom-4   3.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
Range-4       3.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

### DIFF
--- a/it/math.go
+++ b/it/math.go
@@ -12,8 +12,8 @@ import (
 // Range creates a sequence of numbers (positive and/or negative) with given length.
 // Play: https://go.dev/play/p/6ksL0W6KEuQ
 func Range(elementNum int) iter.Seq[int] {
-	length := lo.If(elementNum < 0, -elementNum).Else(elementNum)
-	step := lo.If(elementNum < 0, -1).Else(1)
+	step := lo.Ternary(elementNum < 0, -1, 1)
+	length := elementNum * step
 	return func(yield func(int) bool) {
 		for i, j := 0, 0; i < length; i, j = i+1, j+step {
 			if !yield(j) {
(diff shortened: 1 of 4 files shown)


samber:
i suppose it would be faster in a single condition ?

d-enk:
> i suppose it would be faster in a single condition ?

Well, `length := elementNum * step` is definitely cleaner.

PR #776, opened 2026-01-21
add SliceToSeq2

yangjinheng:
a great method

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/samber/lo
 
-go 1.18
+go 1.25
 
 //
 // Dev dependencies are excluded from releases. Please check CI.
--- a/slice.go
+++ b/slice.go
@@ -1,11 +1,31 @@
 package lo
 
 import (
+	"iter"
 	"sort"
 
 	"github.com/samber/lo/internal/constraints"
 	"github.com/samber/lo/mutable"
 )
+
+// SliceToSeq2 converts a slice into a sequence of pairs (key, value).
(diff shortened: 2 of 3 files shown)


slice.go
```diff
@@ -1,12 +1,32 @@
...
 	"sort"
 
 	"github.com/samber/lo/internal/constraints"
 	"github.com/samber/lo/mutable"
 )
 
+// SliceToSeq2 converts a slice into a sequence of pairs (key, value).
+// Elements are paired consecutively: s[0] with s[1], s[2] with s[3], etc.
+// If the slice has an odd number of elements, the last element is paired with the zero value of type E.
+func SliceToSeq2[Slice ~[]E, E any](collection Slice) iter.Seq2[E, E] {
```
samber, line 14: I would rename it `Pairwise`.

See the `samber/ro` implem: https://ro.samber.dev/docs/operator/combining#pairwise

Please write 2 implementations:
- core package -> returns `[][]T` or `[]T2[T, T]`
- `it/` package -> returns `iter.Seq[[]T]` or `iter.Seq[T2[T, T]]`

Using iter.Seq2 does not seem a good idea here, since first argument of Seq2 is expected to be a key.

I would use `T2` tuples instead of simple slices `[]T`.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/samber/lo
 
-go 1.18
+go 1.25
```
samber, line 3: Please revert. I try to keep a good compatiibility with older go versions.

PR #784, opened 2026-01-22
refactor: introduce `clone` helper for slice copying

d-enk:
- Add clone() function that returns a shallow copy without preserving nilness
  is guaranteed to be inlined by the compiler (verified with -gcflags=-m) 
  (separate from Clone() which preserves nilness - library convention is to return empty slice)
- Simplify Clone() to delegate to clone() for non-nil collections
- Refactor Chunk, Sliding, Drop, DropRight, DropWhile, DropRightWhile, Take, TakeWhile, Replace, Trim to use clone()
- Update SamplesBy and debounce.reset to use clone()
- Replace make(Slice, 0) with Slice{} for return statements

Also
- Some redundant checks have been removed in Take, NthOrEmpty
- Simplify TrimPrefix and TrimSuffix loops
- Micro-optimize Replace() by reordering condition to short-circuit on n != 0 first
  (TODO: consider adding early break when n reaches 0)


d-enk:
Draft reason https://github.com/golang/go/issues/77287

PR #790, opened 2026-01-29
Add two function EqualPtr and ComparePtr

uginroot:
## Summary                                                                                                                                                                                   
                                                                                                                                                                                               
  Add generic pointer comparison functions:                                                                                                                                                    
  - `EqualPtr[T comparable]` — compares two pointers for equality                                                                                                                              
  - `ComparePtr[T constraints.Ordered]` — compares two pointers, returning -1, 0, or 1                                                                                                                 
                                                                                                                                                                                               
  Both functions handle nil pointers safely, treating nil as less than any non-nil value.

### DIFF
--- a/docs/data/core-coalesce.md
+++ b/docs/data/core-coalesce.md
@@ -1,7 +1,7 @@
 ---
 name: Coalesce
 slug: coalesce
-sourceRef: type_manipulation.go#L156
+sourceRef: type_manipulation.go#L157
 category: core
 subCategory: type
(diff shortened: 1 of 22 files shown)


type_manipulation.go
```diff
@@ -1,6 +1,9 @@
 package lo
 
-import "reflect"
+import (
+	"cmp"
```
d-enk: `github.com/samber/lo/internal/constraints`

PR #791, opened 2026-01-30
Add iterator slice helpers

juliazadorozhnaya:
This PR is a [follow‑up](https://github.com/samber/lo/pull/760) to the previous slice‑helpers PR. It adds iterator versions of Take, TakeWhile, and TakeFilter along with tests, examples, and docs.

### DIFF
--- a/docs/static/llms.txt
+++ b/docs/static/llms.txt
@@ -477,6 +477,8 @@ The lo/it package provides iterator helpers for lazy evaluation and streaming op
 - GroupBy: Group elements by key function
 - GroupByMap: Group elements by key function into map
 - Chunk: Split sequence into chunks of specified size
+- Window: Create sliding windows of specified size (overlapping)
+- Sliding: Create sliding windows with specified size and step
 - PartitionBy: Split elements into two groups by predicate
 - Flatten: Flatten nested sequences into single sequence
 - Concat: Combine multiple sequences into one
@@ -498,7 +500,10 @@ The lo/it package provides iterator helpers for lazy evaluation and streaming op
 - DropLast: Remove last n elements
 - DropWhile: Remove elements from start while predicate is true
 - DropLastWhile: Remove elements from end while predicate is true
+- Take: Get first n elements from sequence
+- TakeWhile: Get elements from start while predicate is true
 - DropByIndex: Remove elements at specified indices
+- TakeFilter / TakeFilterI: Filter and take first n matching elements
 - Reject: Remove elements matching predicate
 - RejectI: Remove elements matching index predicate
 - RejectMap: Remove and transform elements matching predicate


it/seq.go
```diff
@@ -624,6 +678,44 @@ func DropLastWhile[T any, I ~func(func(T) bool)](collection I, predicate func(it
...
+	return func(yield func(T) bool) {
+		count := 0
+		for item := range collection {
+			if !yield(item) {
+				return
+			}
+			count++
+			if count >= n {
+				return
+			}
```
d-enk: What if we reduce it to one condition?

Just like it is already in all other places, like:

https://github.com/samber/lo/blob/b6d1bf3749becba82baac0d07db79b6c2255382e/it/seq.go#L575-L579

channel.go
```diff
@@ -325,3 +325,69 @@ func FanOut[T any](count, channelsBufferCap int, upstream <-chan T) []<-chan T {
 
 	return channelsToReadOnly(downstreams)
 }
+
+// Distinct returns a channel with duplicate values removed.
+// The first occurrence is preserved, and ordering is maintained.
+func Distinct[T comparable](upstream <-chan T) <-chan T {
```
d-enk: It should at least be mentioned that memory can run out.

channel.go
```diff
@@ -325,3 +325,69 @@ func FanOut[T any](count, channelsBufferCap int, upstream <-chan T) []<-chan T {
...
+			unique <- value
+		}
+	}()
+
+	return unique
+}
+
+// Tee duplicates the stream into multiple output channels without blocking.
+// If an output channel is full or not ready, the value is dropped for that channel.
+func Tee[T any](count, channelsBufferCap int, upstream <-chan T) []<-chan T {
```
d-enk: In my opinion, this function is too specific.

it/seq.go
juliazadorozhnaya: Done - condensed to a single condition as suggested. Thanks

channel.go
juliazadorozhnaya: Added a note in the doc comments

channel.go
juliazadorozhnaya: Do you mean we should remove Tee because FanOut/FanIn already cover this, or keep it as a separate helper?

it/seq.go
```diff
@@ -691,11 +691,8 @@ func Take[T any, I ~func(func(T) bool)](collection I, n int) I {
 	return func(yield func(T) bool) {
 		count := 0
 		for item := range collection {
-			if !yield(item) {
-				return
-			}
 			count++
-			if count >= n {
+			if count > n || !yield(item) {
```
d-enk: In your case you need to avoid additional iteration by collection

```go
if !yield(item) || count >= n  {
```

Sorry for the misleading comment

d-enk: I meant all the relevant places

`TakeWhile`, `TakeFilter` too

it/seq.go
```diff
@@ -312,6 +312,60 @@ func Chunk[T any](collection iter.Seq[T], size int) iter.Seq[[]T] {
...
+			if !yield(window) {
+				return
+			}
+
+			if step >= size {
+				buffer = buffer[:0]
+				skip = step - size
+			} else {
+				buffer = buffer[step:]
+			}
```
d-enk: `step`, `size` is constant

So this may be precalculated before loop

juliazadorozhnaya: done

it/seq.go
juliazadorozhnaya: done

it/seq.go
d-enk: This may be simplified to something like

```go
skip = skipDelta
buffer = buffer[I:J]
```

But now I've actually noticed that this forward buffer shift is incorrect.

This leads to expansion and reallocation on next appends

juliazadorozhnaya: Good catch, thanks. Updated the forward shift to keep capacity by copying the overlap to the front and reslicing so we avoid extra reallocations on subsequent appends.

juliazadorozhnaya: Managed to improve the implementation, benchmarks show about a 10% speedup

```
BenchmarkItSliding/old_window_step1_n10
BenchmarkItSliding/old_window_step1_n10-11         	 2913607	       379.6 ns/op
BenchmarkItSliding/new_window_step1_n10
BenchmarkItSliding/new_window_step1_n10-11         	 3341169	       341.3 ns/op
BenchmarkItSliding/old_overlap_step3_n10
BenchmarkItSliding/old_overlap_step3_n10-11        	 4146115	       285.8 ns/op
BenchmarkItSliding/new_overlap_step3_n10
BenchmarkItSliding/new_overlap_step3_n10-11        	 4664964	       250.7 ns/op
BenchmarkItSliding/old_no_overlap_n10
BenchmarkItSliding/old_no_overlap_n10-11           	 4272240	       280.9 ns/op
BenchmarkItSliding/new_no_overlap_n10
BenchmarkItSliding/new_no_overlap_n10-11           	 4828736	       249.4 ns/op
BenchmarkItSliding/old_gap_step8_n10
BenchmarkItSliding/old_gap_step8_n10-11            	 4425630	       263.0 ns/op
BenchmarkItSliding/new_gap_step8_n10
BenchmarkItSliding/new_gap_step8_n10-11            	 5213691	       235.1 ns/op
BenchmarkItSliding/old_window_step1_n100
BenchmarkItSliding/old_window_step1_n100-11        	  382444	      3064 ns/op
BenchmarkItSliding/new_window_step1_n100
BenchmarkItSliding/new_window_step1_n100-11        	  406326	      2968 ns/op
BenchmarkItSliding/old_overlap_step3_n100
BenchmarkItSliding/old_overlap_step3_n100-11       	  652431	      1794 ns/op
BenchmarkItSliding/new_overlap_step3_n100
BenchmarkItSliding/new_overlap_step3_n100-11       	  703446	      1694 ns/op
BenchmarkItSliding/old_no_overlap_n100
BenchmarkItSliding/old_no_overlap_n100-11          	  778430	      1523 ns/op
BenchmarkItSliding/new_no_overlap_n100
BenchmarkItSliding/new_no_overlap_n100-11          	  812672	      1454 ns/op
BenchmarkItSliding/old_gap_step8_n100
BenchmarkItSliding/old_gap_step8_n100-11           	  855152	      1381 ns/op
BenchmarkItSliding/new_gap_step8_n100
BenchmarkItSliding/new_gap_step8_n100-11           	  914552	      1303 ns/op
BenchmarkItSliding/old_window_step1_n1000
BenchmarkItSliding/old_window_step1_n1000-11       	   40515	     29331 ns/op
BenchmarkItSliding/new_window_step1_n1000
BenchmarkItSliding/new_window_step1_n1000-11       	   42220	     28129 ns/op
BenchmarkItSliding/old_overlap_step3_n1000
BenchmarkItSliding/old_overlap_step3_n1000-11      	   71319	     16808 ns/op
BenchmarkItSliding/new_overlap_step3_n1000
BenchmarkItSliding/new_overlap_step3_n1000-11      	   74110	     16148 ns/op
BenchmarkItSliding/old_no_overlap_n1000
BenchmarkItSliding/old_no_overlap_n1000-11         	   86786	     13783 ns/op
BenchmarkItSliding/new_no_overlap_n1000
BenchmarkItSliding/new_no_overlap_n1000-11         	   89229	     13398 ns/op
BenchmarkItSliding/old_gap_step8_n1000
BenchmarkItSliding/old_gap_step8_n1000-11          	   94996	     12489 ns/op
BenchmarkItSliding/new_gap_step8_n1000
BenchmarkItSliding/new_gap_step8_n1000-11          	   99624	     11980 ns/op
```

d-enk:
I still find channel functions questionable.

I suggest splitting the pull request into two parts.

As it should have been from the beginning

d-enk:
Just as I already suggested in a private message, I propose to split `Sliding` explicitly

```go
switch {
case step == size:
...
case step < size:
...
case step > size:
...
}
```

it/seq.go
```diff
@@ -632,6 +722,32 @@ func DropByIndex[T any, I ~func(func(T) bool)](collection I, indexes ...int) I {
 	return RejectI(collection, func(_ T, index int) bool { return lo.HasKey(set, index) })
 }
 
+// TakeFilter filters elements and takes the first n elements that match the predicate.
+// Equivalent to calling Take(FilterI(...)), but more efficient as it stops after finding n matches.
+func TakeFilter[T any, I ~func(func(T) bool)](collection I, n int, predicate func(item T, index int) bool) I {
```
samber: Can you rename it to `TakeFilterI` and add a `TakeFilter` variant please?

(also in error below + doc)

channel.go
samber: I agree with @d-enk .

Please open a new PR for further discussion.

Also: I would rename it in something more similar to `FanOut`: `FanOutBestEffort`, `FanOutNonBlocking`, `FanOutLossy`, `FanOutDrop`, `FanOutFireAndForget`, other ?

it/seq.go
juliazadorozhnaya: Done. Renamed to TakeFilterI and added a TakeFilter variant.

channel.go
juliazadorozhnaya: Done. I've removed the Tee function and all references to it from this PR.

juliazadorozhnaya:
> Just as I already suggested in a private message, I propose to split `Sliding` explicitly
> 
> ```go
> switch {
> case step == size:
> ...
> case step < size:
> ...
> case step > size:
> ...
> }
> ```

Done. Sliding now uses an explicit switch on offset (step - size) for the three cases.

d-enk:
I'd remove Distinct/DistinctBy too.

You can also clean up the title/commits.

d-enk:
> Can you rename it to TakeFilterI and add a TakeFilter variant please?

@samber Something went wrong somewhere

https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L12

https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L653

https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L713

juliazadorozhnaya:
> > Can you rename it to TakeFilterI and add a TakeFilter variant please?
> 
> Something went wrong somewhere
> 
> https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L12
> 
> https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L653
> 
> https://github.com/samber/lo/blob/921fb61893343d3b67b927df9c9894b962d39049/slice.go#L713

Done. In slice.go renamed to TakeFilterI and added TakeFilter variant

juliazadorozhnaya:
> I'd remove Distinct/DistinctBy too.
> 
> You can also clean up the title/commits.

Done

d-enk:
> Done. In slice.go renamed to TakeFilterI and added TakeFilter variant


https://github.com/samber/lo/pull/791#issuecomment-3945399638

This was not a call to action for you...

samber:
Oops, sorry. You wrote `Take(FilterI(...))` so that why i suggested `TakeFilterI` but lets use `TakeFilter` instead. Sorry for the mistake.

EDIT: I reverted the last commit.

EDIT2: Thanks to both of you. I'm targeting a new release this week.

d-enk:
I completely missed that ```Take(Filter())``` already exists

As if ```TakeFilter()``` were redundant for sequences...

PR #796, opened 2026-02-10
fix: make Ellipsis operate on runes instead of bytes to prevent Unicode truncation

### LINKED ISSUE #520: Ellipsis: avoid Unicode character truncation
While you are working on Ellipsis, please look at the code behavior.

I doubt the code does what is expected with a Unicode character like 🧜‍♂️ by truncating to Len 1

veeceey:
## Summary

- **Fixes #520**: `Ellipsis` previously used byte-based length counting (`len(str)`) and byte-based slicing (`str[:length-3]`), which could split multi-byte Unicode characters (emoji, CJK, accented letters) in the middle, producing garbled output.
- Changed `Ellipsis` to use `[]rune` conversion so the `length` parameter counts Unicode code points (runes) instead of bytes. Multi-byte characters are now never split.
- Added comprehensive Unicode test cases covering CJK characters, emoji, and accented characters.
- Updated documentation in `README.md`, `string.go` godoc, and `docs/data/core-ellipsis.md`.

## Details

The core change is minimal (3 lines in `string.go`):

```go
// Before (byte-based, breaks Unicode):
if len(str) > length {
    return strings.TrimSpace(str[:length-3]) + "..."
}

// After (rune-based, Unicode-safe):
runes := []rune(str)
if len(runes) > length {
    return strings.TrimSpace(string(runes[:length-3])) + "..."
}
```

**Note on grapheme clusters**: This fix addresses rune-level truncation. Complex grapheme clusters composed of multiple code points (e.g., `🧜‍♂️` which uses zero-width joiners) may still be split between their constituent runes. Full grapheme segmentation would require either an external dependency or a substantial internal implementation, which the maintainer indicated should be avoided. The rune-based approach fixes the vast majority of real-world Unicode truncation issues.

## Test plan

- [x] All existing ASCII-based `TestEllipsis` tests continue to pass (rune count == byte count for ASCII)
- [x] New test cases verify correct behavior with CJK characters (`世界`, `你好`)
- [x] New test cases verify correct behavior with emoji (`🏠🐶🐱🌟🎉🌈`)
- [x] New test cases verify correct behavior with accented characters (`café`)
- [x] Full test suite passes: `go test ./...`

### DIFF
--- a/string.go
+++ b/string.go
@@ -235,16 +235,6 @@ func Capitalize(str string) string {
 func Ellipsis(str string, length int) string {
 	str = strings.TrimSpace(str)
 
-	if length < 3 {
-		// Can't fit any content before the ellipsis.
-		for range str {
-			if length--; length < 0 {
-				return "..."
-			}
-		}
-		return str
-	}
-
 	const ellipsis = "..."
 
 	cutPosition := 0
@@ -252,6 +242,7 @@ func Ellipsis(str string, length int) string {
 		if length == len(ellipsis) {
 			cutPosition = i
 		}
+
 		if length--; length < 0 {
 			return strings.TrimSpace(str[:cutPosition]) + ellipsis
 		}


d-enk:
Optimized code (without convertions)

```go
func Ellipsis(str string, length int) string {
	str = strings.TrimSpace(str)

	const ellipsis = "..."

	cutPosition := 0
	for i := range str {
		if length == len(ellipsis) {
			cutPosition = i
		}

		if length--; length < 0 {
			return strings.TrimSpace(str[:cutPosition]) + ellipsis
		}
	}

	return str
}
```

veeceey:
Nice optimization\! Using `range` to iterate runes avoids the `[]rune(str)` allocation entirely. I've adopted your approach with one tweak: added an early return for the `length < 3` edge case before the loop to keep the intent explicit, rather than relying on `cutPosition` staying at 0. Updated the PR -- thanks for the suggestion\!

d-enk:
In my opinion

An early return that almost completely duplicates the main algorithm 
for the sake of an extremely rare use case seems unnecessary.

I even abandoned the check `if cutPosition == 0 {}`

however, at least reuse the `const ellipsis = "..."`

veeceey:
Fair point -- the early return does duplicate logic for a rare edge case. I've simplified it: removed the early return and just use the `ellipsis` const for the fallback. Cleaner this way, thanks for the feedback.

veeceey:
Hi @samber, just a gentle ping on this one — I've incorporated the optimized rune iteration approach that @d-enk suggested and simplified the edge case handling. The fix prevents Unicode truncation in Ellipsis by operating on runes instead of bytes. Would love your review\!

samber:
Thanks guys, for the work on this 👏

ccoVeille:
I had surprise by using runes


Here the code won't do what is expected with Unicode character made of multiple Unicode char

https://go.dev/play/p/s-pb7mlpQVj

ccoVeille:
As mentioned here:

> This bug could be fixed by counting the string grapheme cluster at `github.com/rivo/uniseg` .
>
> If this policy is acceptable, I'll work on a PR for the fix!
> 
_Originally posted by @yasuaki640 in https://github.com/samber/lo/issues/520#issuecomment-2615682710_

Using @rivo uniseg  could be a solution. But as stated before adding a dependency is not recorded by @samber, so here checking what uniseg code do abd replicate it could be a solution

samber:
Another option would be to add `github.com/rivo/uniseg` in the comment of the helper for advanced cases.

@ccoVeille i'm going to add your example as a commented test case. If you want to make a PR to try to fix it, please do !

PR #799, opened 2026-02-15
fix: reset n counter per iteration in it.Replace

### LINKED ISSUE #798: Question: n counter in it.Replace is shared across iterations
`it.Replace` captures `n`, so `n--` persists across iterations ⇒ second time you iterate the sequence, replacements don't happen :(

```go
package main

import (
	"fmt"
	"slices"

	"github.com/samber/lo/it"
)

func main() {
	seq := it.Replace(slices.Values([]int{1, 2, 1, 2, 1}), 1, 9, 2)

	fmt.Println(slices.Collect(seq)) // expected: [9 2 9 2 1] actual: [9 2 9 2 1]
	fmt.Println(slices.Collect(seq)) // expected: [9 2 9 2 1] actual: [1 2 1 2 1]
}
```

Other `it` functions like `Uniq` create fresh state per iteration, so this looks unintentional. Real impact is small because most users collect the sequence once, is this considered a bug?

LikimiaD:
Fixes #798

### DIFF
--- a/it/seq.go
+++ b/it/seq.go
@@ -760,13 +760,16 @@ func Slice[T any, I ~func(func(T) bool)](collection I, start, end int) I {
 // Replace returns a sequence with the first n non-overlapping instances of old replaced by new.
 // Play: https://go.dev/play/p/6XrKO0-zw
 func Replace[T comparable, I ~func(func(T) bool)](collection I, old, nEw T, n int) I {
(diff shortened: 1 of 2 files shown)


LikimiaD:
Сhecked сhanges - seems that minimal edits were needed to fix the bug

samber:
Thanks !

PR #801, opened 2026-02-16
Experiments: adding SIMD helpers

samber:
Introducing `github.com/samber/lo/exp/simd` package.

Go 1.26 brings SIMD instructions to Go.

The feature is still experimental, so i'm not adding this to the core package of `lo`.

```go
# usage
import losimd "github.com/samber/lo/exp/simd"

result := losimd.SumInt8x64([]int8{1, 2, 3, ...})
```

Requirements:
- amd64 architecture
- AVX2 or AVX512 CPU
- build with `GOEXPERIMENT=simd`

I've added `SumInt8` variants that route the implementation to the right architecture automatically. I would be interested by your opinion on this.

Should we support vaargs ? If yes, should `losimd.SumInt8x64(arg1, arg2, arg3)` returns a slice of 3 results or a single result?

@jorropo @d-enk 

Warning: The experimental nature of this sub-package implies that we might have breaking changes in the future.

TODO:
- [x] Benchmarks
- [x] Documentation: it is an experimental feature, and we might break the API in the future

### DIFF
--- a/exp/simd/BENCHMARK.md
+++ b/exp/simd/BENCHMARK.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Benchmarks show that running SIMD operators on small datasets is slower:
+Benchmarks show that running SIMD operations on small datasets is slower:
 
 ```txt
 BenchmarkSumInt8/small/Fallback-lo-2            	248740710	        5.218 ns/op
@@ -11,7 +11,7 @@ BenchmarkSumInt8/small/AVX2-x32-2               	 73059427	        14.44 ns/op
 BenchmarkSumInt8/small/AVX512-x64-2             	 49913169	        24.41 ns/op
 ```
 
-But much much faster on big datasets:
+But SIMD is much faster on large datasets:
 
 ```txt
 BenchmarkSumInt8/xlarge/Fallback-lo-2           	  273898	         4383 ns/op


exp/simd/intersect_sse_test.go
```diff
@@ -0,0 +1,590 @@
...
+
+import (
+	"math/rand/v2"
+	"testing"
+
+	"github.com/samber/lo"
+)
+
+func TestContainsInt8x16(t *testing.T) {
+	requireAVX512(t)
```
d-enk: Wrong requireAVX512 for sse

exp/simd/math.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/samber/lo"
+)
+
+func SumInt8[T ~int8](collection []T) T {
+	if archsimd.X86.AVX512() {
+		return SumInt8x64(collection)
+	} else if archsimd.X86.AVX2() {
+		return SumInt8x32(collection)
+	}
+	return lo.Sum(collection)
```
d-enk: Need SSE fallback `return SumInt8x16(collection)`

exp/simd/intersect_sse.go
```diff
@@ -0,0 +1,340 @@
+//go:build go1.26 && goexperiment.simd && amd64
+
+package simd
+
+import (
+	"simd/archsimd"
+	"unsafe"
+)
+
+// ContainsInt8x16 checks if collection contains target using SSE SIMD and AVX512 SIMD
```
d-enk: AVX512 is not used

exp/simd/math_avx2.go
```diff
@@ -0,0 +1,1820 @@
...
+
+	i := 0
+	for ; i+lanes <= len(collection); i += lanes {
+		v := archsimd.LoadInt8x32Slice(base[i : i+lanes])
+
+		clamped := v.Max(minVec).Min(maxVec)
+
+		var buf [32]int8
+		clamped.Store(&buf)
+		for j := 0; j < lanes; j++ {
```
d-enk:
```go
copy(baseResult[i:], buf[:])
```

d-enk:
```go
dst := (*[32]int8)(unsafe.Pointer(&result[i]))
clamped.Store(dst)
```

```go
clamped.Store((*[32]int8)(baseResult[i:]))
```

exp/simd/intersect_sse_test.go
samber: It uses `mask.ToBits()` which requires avx512.

We could imagine 2 conditional branches: AVX512 and non-AVX512?

d-enk: I didn't really look into it

The test just worked successfully for me.

d-enk: Only `TestContainsUint16x8, TestContainsUint16x8` not works by `illegal instruction`

samber: Can you copy your /proc/cpuinfo please?

Since I code on MacOS, I need to start a VM to test all architectures.  🥵

Having AVX512 instructions but not AVX2 seems weired.

exp/simd/intersect_avx512.go
```diff
@@ -0,0 +1,969 @@
+//go:build go1.26 && goexperiment.simd && amd64
+
+package simd
+
+import (
+	"simd/archsimd"
+)
+
+// ContainsInt8x16 checks if collection contains target using SSE SIMD and AVX512 SIMD
```
d-enk, line 9: AVX2

exp/simd/intersect_sse_test.go
d-enk: archsimd.X86Features

```yaml
HasAES: true
HasADX: true
HasAVX: true
HasAVXVNNI: false
HasAVX2: true
HasAVX512: false
HasAVX512F: false
HasAVX512CD: false
HasAVX512BW: false
HasAVX512DQ: false
HasAVX512VL: false
HasAVX512GFNI: false
HasAVX512VAES: false
HasAVX512VNNI: false
HasAVX512VBMI: false
HasAVX512VBMI2: false
HasAVX512BITALG: false
HasAVX512VPOPCNTDQ: false
HasAVX512VPCLMULQDQ: false
HasBMI1: true
HasBMI2: true
HasERMS: true
HasFSRM: false
HasFMA: true
HasGFNI: false
HasOSXSAVE: true
HasPCLMULQDQ: true
HasPOPCNT: true
HasRDTSCP: true
HasSHA: false
HasSSE3: true
HasSSSE3: true
HasSSE41: true
HasSSE42: true
HasVAES: false
```

d-enk:
You can also use for optimization:
- const lanes
- uint index

For example, it became ~30% faster.

```go
func SumInt16x8[T ~int16](collection []T) T {
	length := uint(len(collection))
	if length == 0 {
		return 0
	}

	const lanes = 8

	base := unsafeSliceInt16(collection, int(length))
	var acc archsimd.Int16x8

	i := uint(0)
	for ; i+lanes <= length; i += lanes {
		v := archsimd.LoadInt16x8Slice(base[i : i+lanes])
		acc = acc.Add(v)
	}

	var buf [lanes]int16
	acc.Store(&buf)
	var sum T
	for k := range uint(lanes) {
		sum += T(buf[k])
	}

	for ; i < length; i++ {
		sum += collection[i]
	}

	return sum
}
```

exp/simd/intersect_sse_test.go
samber: i'm pretty sure to have fixed all architecture checks

Can you run tests again please?

For CI, I moved to ubicloud, which supports AVX512. Github seems to have multiple hardware configs and none supports AVX512 during the weekend :-(

PR #824, opened 2026-03-02
feat: support for buffer iterator

mimol91:
Add support for Buffer in `it` sub package

### DIFF
--- a/docs/CLAUDE.md
+++ b/docs/CLAUDE.md
@@ -88,6 +88,7 @@ Use `variantHelpers` for different versions of the **same helper function**:
 ```yaml
 variantHelpers:
   - core#slice#map        # func Map[T, R]([]T, func(T, int) R) []R
+  - core#slice#maperr     # func MapErr[T, R]([]T, func(T, int) (R, error)) ([]R, error)
   - core#slice#mapi       # func MapI[T, R]([]T, func(T, int) R) []R (with index)
   - core#slice#mapwithcontext  # func MapWithContext[T, R]([]T, func(T, int, context.Context) R, context.Context) []R
 ```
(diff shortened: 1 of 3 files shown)


it/channel.go
```diff
@@ -51,3 +51,28 @@ func ChannelToSeq[T any](ch <-chan T) iter.Seq[T] {
...
+func Buffer[T any](ch <-chan T, size int) iter.Seq[[]T] {
+	return func(yield func([]T) bool) {
+		for {
+			buffer := make([]T, 0, size)
+
+			for range size {
+				item, ok := <-ch
+				if !ok {
+					if len(buffer) > 0 {
+						yield(buffer)
```
d-enk: ```_ = yield(buffer)```

mimol91: Could you explain why this change? 
Value is not used, and function does not return error

d-enk: This makes it clearer that this isn't just a forgotten check.

Personally, I thought exactly that at first glance.

samber:
I would change the control flow: channel is push-based vs iterators are pull-based.

```go
func Buffer[T any](seq iter.Seq[T], size int) iter.Seq[[]T] {
      return func(yield func([]T) bool) {
            buffer := make([]T, 0, size)

            seq(func(v T) bool {
                  buffer = append(buffer, v)
                  if len(buffer) < size {
                      return true  // keep pulling
                  }
                  // Buffer full, yield it
                  result := buffer
                  buffer = make([]T, 0, size)  // allocate new buffer
                  return yield(result)  // false = stop, true = continue
            })

            // Yield remaining partial buffer
            if len(buffer) > 0 {
                yield(buffer)
            }
      }
}
```

WDYT?

EDIT: `SeqToChannel` could make the transformation from channel to seq.

EDIT2: i'm doing the change by myself since i would like to release the next version of `lo` today

PR #858, opened 2026-04-02
feat: add concurrency limit and error handling to parallel package

GiGurra:
## Summary

Adds bounded concurrency and error handling to the `parallel` package, addressing a long-standing request:

- **`WithConcurrency(n)` option** added to all existing parallel functions (`Map`, `ForEach`, `Times`, `GroupBy`, `PartitionBy`) — fully backward compatible
- **`MapErr`** and **`ForEachErr`** — new variants with error propagation (returns first error, stops processing) and `WithContext` support for cancellation

### Usage

```go
// Bounded concurrency (backward compatible — existing calls work unchanged)
parallel.Map(urls, fetch, parallel.WithConcurrency(10))
parallel.ForEach(items, process, parallel.WithConcurrency(5))

// Error handling + concurrency + context cancellation
results, err := parallel.MapErr(urls, fetch, parallel.WithConcurrency(10), parallel.WithContext(ctx))
err := parallel.ForEachErr(items, process, parallel.WithConcurrency(5))
```

### Context

This addresses #77 and supersedes the stalled #37 (approved in 2022 but never merged due to conflicts). The approach is simpler: functional options instead of a task pool struct, and a proper bounded worker pool.

For a reference implementation of a more full-featured parallel processing library (recursive-safe worker pools, etc.), see [github.com/GiGurra/party](https://github.com/GiGurra/party).

### Implementation notes

- **Bounded worker pool**: `WithConcurrency(n)` spins up exactly `min(n, len(items))` worker goroutines that pull work indices from an unbuffered channel — no more goroutines than requested.
- Without `WithConcurrency`, the original one-goroutine-per-item behavior is preserved (fully backward compatible).
- **Error handling via channel**: Workers push errors to a 1-buffered channel (first error wins). The sender `select`s on work dispatch, error channel, and context cancellation in a single statement — no atomics or sync.Once needed.
- On error or context cancellation, the sender stops dispatching, closes the work channel, and waits for in-flight workers to finish before returning.
- All existing tests pass, new tests cover: bounded concurrency verification, error propagation, context cancellation, empty collections, edge cases, and a data race regression test.
- Full test suite passes with `-race`

### Future enhancements

- **`WithWorkerPool(pool)`**: Allow reusing a pre-allocated pool across multiple calls, avoiding per-call setup overhead for hot paths.
- **Configurable error handling**: Currently the Err variants fail fast, returning the first error. A future option like `WithCollectErrors()` could instead collect all errors and return a joined/composite error (`errors.Join`), letting all items run to completion — or stop after a configurable error threshold (e.g. `WithMaxErrors(n)`).

The functional options pattern makes both of these straightforward to add without breaking changes.

### DIFF
--- a/parallel/bench_test.go
+++ b/parallel/bench_test.go
@@ -2,8 +2,62 @@ package parallel
 
 import (
 	"context"
+	"sync"
 	"testing"
 )
+
+// Upstream master implementations inlined for direct comparison.
+// Run `go test -bench=. ./parallel/` to see them side by side.
+
+func masterMap(items []int, transform func(int, int) int) []int {
+	result := make([]int, len(items))
+	var wg sync.WaitGroup
+	wg.Add(len(items))
+	for i, item := range items {
+		go func(_item, _i int) {
+			result[_i] = transform(_item, _i)
+			wg.Done()
+		}(item, i)
+	}
+	wg.Wait()
+	return result
+}
+
+func masterForEach(items []int, callback func(int, int)) {
+	var wg sync.WaitGroup
+	wg.Add(len(items))
+	for i, item := range items {
+		go func(_item, _i int) {
+			callback(_item, _i)
+			wg.Done()
+		}(item, i)
+	}
+	wg.Wait()
+}
+
+func masterTimes(count int, iteratee func(int) int) []int {
+	result := make([]int, count)
+	var wg sync.WaitGroup
+	wg.Add(count)
+	for i := 0; i < count; i++ {
+		go func(_i int) {
+			result[_i] = iteratee(_i)
+			wg.Done()
+		}(i)
+	}
+	wg.Wait()
+	return result
+}
+
+func BenchmarkMap_Master(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		masterMap(items, func(x, _ int) int { return x * 2 })
+	}
+}
 
 func BenchmarkMap(b *testing.B) {
 	items := make([]int, 100)
@@ -15,6 +69,16 @@ func BenchmarkMap(b *testing.B) {
 	}
 }
 
+func BenchmarkForEach_Master(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		masterForEach(items, func(_, _ int) {})
+	}
+}
+
 func BenchmarkForEach(b *testing.B) {
 	items := make([]int, 100)
 	for i := range items {
@@ -22,6 +86,12 @@ func BenchmarkForEach(b *testing.B) {
 	}
 	for b.Loop() {
 		ForEach(items, func(_, _ int) {})
+	}
+}
+
+func BenchmarkTimes_Master(b *testing.B) {
+	for b.Loop() {
+		masterTimes(100, func(i int) int { return i * 2 })
 	}
 }
 
--- a/parallel/bench_test.go
+++ b/parallel/bench_test.go
@@ -1,6 +1,9 @@
 package parallel
 
-import "testing"
+import (
+	"context"
+	"testing"
+)
 
 func BenchmarkMap(b *testing.B) {
 	items := make([]int, 100)
@@ -25,5 +28,36 @@ func BenchmarkForEach(b *testing.B) {
 func BenchmarkTimes(b *testing.B) {
 	for b.Loop() {
 		Times(100, func(i int) int { return i * 2 })
+	}
+}
+
+func BenchmarkMapErr_Unbounded(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		_, _ = MapErr(items, func(x, _ int) (int, error) { return x * 2, nil })
+	}
+}
+
+func BenchmarkMapErr_UnboundedWithContext(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	ctx := context.Background()
+	for b.Loop() {
+		_, _ = MapErr(items, func(x, _ int) (int, error) { return x * 2, nil }, WithContext(ctx))
+	}
+}
+
+func BenchmarkMapErr_Bounded(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		_, _ = MapErr(items, func(x, _ int) (int, error) { return x * 2, nil }, WithConcurrency(10))
 	}
 }
--- a/parallel/bench_test.go
+++ b/parallel/bench_test.go
@@ -1,0 +1,29 @@
+package parallel
+
+import "testing"
+
+func BenchmarkMap(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		Map(items, func(x, _ int) int { return x * 2 })
+	}
+}
+
+func BenchmarkForEach(b *testing.B) {
+	items := make([]int, 100)
+	for i := range items {
+		items[i] = i
+	}
+	for b.Loop() {
+		ForEach(items, func(_, _ int) {})
+	}
+}
+
+func BenchmarkTimes(b *testing.B) {
+	for b.Loop() {
+		Times(100, func(i int) int { return i * 2 })
+	}
+}
--- a/parallel/options.go
+++ b/parallel/options.go
@@ -28,7 +28,7 @@ func WithContext(ctx context.Context) Option {
 }
 
 func buildOptions(opts []Option) options {
-	o := options{}
+	o := options{ctx: context.Background()}
 	for _, opt := range opts {
 		opt(&o)
 	}
--- a/parallel/options.go
+++ b/parallel/options.go
@@ -7,30 +7,46 @@ type options struct {
 	ctx         context.Context
 }
 
-// Option configures parallel execution.
-type Option func(*options)
+// ErrOption configures Err-variant parallel functions (MapErr, ForEachErr).
+type ErrOption interface {
+	apply(*options)
+}
+
+// Option configures all parallel functions. It extends ErrOption with a
+// cannotFail marker, meaning the option does not require error reporting
+// to work correctly. Options like WithContext that need an error return
+// only implement ErrOption and will be rejected at compile time by
+// non-Err functions.
+type Option interface {
+	ErrOption
+	cannotFail()
+}
+
+// ConcurrencyOpt is the option returned by WithConcurrency.
+type ConcurrencyOpt int
+
+func (c ConcurrencyOpt) apply(o *options) { o.concurrency = int(c) }
+func (c ConcurrencyOpt) cannotFail()      {}
 
 // WithConcurrency sets the maximum number of goroutines running at the same time.
 // A value <= 0 means unbounded (one goroutine per item).
-func WithConcurrency(n int) Option {
-	return func(o *options) {
-		o.concurrency = n
-	}
-}
+func WithConcurrency(n int) ConcurrencyOpt { return ConcurrencyOpt(n) }
+
+// ContextOpt is the option returned by WithContext.
+type ContextOpt struct{ ctx context.Context }
+
+func (c ContextOpt) apply(o *options) { o.ctx = c.ctx }
 
 // WithContext sets a context for cancellation support.
 // When the context is cancelled, no new items are processed and the first
-// cancellation error is returned. Only applies to Err variants.
-func WithContext(ctx context.Context) Option {
-	return func(o *options) {
-		o.ctx = ctx
-	}
-}
+// cancellation error is returned. Only usable with Err variants (MapErr,
+// ForEachErr); passing it to non-Err functions is a compile error.
+func WithContext(ctx context.Context) ContextOpt { return ContextOpt{ctx: ctx} }
 
-func buildOptions(opts []Option) options {
+func buildOptions[O ErrOption](opts []O) options {
 	o := options{ctx: context.Background()}
 	for _, opt := range opts {
-		opt(&o)
+		opt.apply(&o)
 	}
 	return o
 }
--- a/parallel/options.go
+++ b/parallel/options.go
@@ -1,0 +1,36 @@
+package parallel
+
+import "context"
+
+type options struct {
+	concurrency int
+	ctx         context.Context
+}
+
+// Option configures parallel execution.
+type Option func(*options)
+
+// WithConcurrency sets the maximum number of goroutines running at the same time.
+// A value <= 0 means unbounded (one goroutine per item).
+func WithConcurrency(n int) Option {
+	return func(o *options) {
+		o.concurrency = n
+	}
+}
+
+// WithContext sets a context for cancellation support.
+// When the context is cancelled, no new items are processed and the first
+// cancellation error is returned. Only applies to Err variants.
+func WithContext(ctx context.Context) Option {
+	return func(o *options) {
+		o.ctx = ctx
+	}
+}
+
+func buildOptions(opts []Option) options {
+	o := options{}
+	for _, opt := range opts {
+		opt(&o)
+	}
+	return o
+}
--- a/parallel/slice.go
+++ b/parallel/slice.go
@@ -1,10 +1,6 @@
 package parallel
 
-import (
-	"context"
-	"sync"
-	"sync/atomic"
-)
+import "sync"
 
 // Map manipulates a slice and transforms it to a slice of another type.
 // `transform` is called in parallel. Result keep the same order.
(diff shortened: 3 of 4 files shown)


parallel/slice.go
```diff
@@ -112,3 +124,100 @@ func PartitionBy[T any, K comparable, Slice ~[]T](collection Slice, iteratee fun
...
+		if atomic.LoadInt32(&hasErr) != 0 {
+			break
+		}
+		if !sendWork(o.ctx, work, i) {
+			setErr(o.ctx.Err())
+			break
+		}
+	}
+
+	return firstErr
```
d-enk: Incorrect with DATA RACE

```go
err := runErr(1, func(i int) error {
		time.Sleep(time.Second)

		return errors.New("Ooops")
	}, options{})

fmt.Println(err)
```

Must be before `return firstErr`

```go
close(work)
wg.Wait()
```

parallel/slice.go
```diff
@@ -112,3 +124,100 @@ func PartitionBy[T any, K comparable, Slice ~[]T](collection Slice, iteratee fun
...
+	wg.Add(workers)
+	defer wg.Wait()
+	defer close(work)
+
+	for w := 0; w < workers; w++ {
+		go func() {
+			defer wg.Done()
+			for i := range work {
+				if atomic.LoadInt32(&hasErr) != 0 {
+					continue
```
d-enk: return

parallel/slice.go
```diff
@@ -112,3 +124,100 @@ func PartitionBy[T any, K comparable, Slice ~[]T](collection Slice, iteratee fun
...
+	)
+
+	setErr := func(err error) {
+		once.Do(func() {
+			firstErr = err
+			atomic.StoreInt32(&hasErr, 1)
+		})
+	}
+
+	workers := minInt(concurrency, n)
```
d-enk, line 213: `concurrency` calculation logic in 2 places

parallel/slice.go
GiGurra: Good catch — you're absolutely right. The deferred `wg.Wait()` ran after `return firstErr`, so a worker could still be writing `firstErr` via `sync.Once` while the caller read it. Fixed in cebe42e by moving `close(work)` and `wg.Wait()` before the return. Also added a regression test (`TestMapErrRaceOnReturn`) that reliably triggers the race under `-race`. Thanks!

parallel/slice.go
GiGurra: ~~This needs to be `continue`, not `return`.~~

Correction: with an unbuffered channel, `return` would also be safe — `for i := range work` receives an item before evaluating the loop body, so a worker always drains one pending send before exiting. The sender then advances, checks `hasErr`, and breaks.

That said, `continue` is slightly more defensive (would still work if the channel were ever changed to buffered), and the cost of draining a few extra items is negligible. I suggest we keep `continue` for now.

parallel/slice.go
GiGurra: Good point — this was the case in an earlier revision. It's now consolidated: the concurrency defaulting (`if concurrency <= 0`) only lives in `runErr`. All callers just pass `buildOptions(opts)` through without touching concurrency.

parallel/slice.go
d-enk: Possibly consider removing the hasErr check - with an unbuffered channel

It adds at most one extra item after an error while eliminating atomic operations per iteration

GiGurra: That's a fair point — with an unbuffered channel the sender can only be one send ahead, so at most one extra item runs after an error. Removing the check would simplify the worker loop and eliminate the per-item atomic load.

The tradeoff is that the sender-side `hasErr` check (which stops dispatching) still does the heavy lifting. The worker-side check is just an optimization to skip work on items already in flight. For cheap callbacks it barely matters; for expensive ones it avoids one unnecessary execution.

Happy to go either way — what do you think?

GiGurra:
Pushed a simplification based on the review feedback — replaced the `sync.Once` + `atomic.Int32` + `hasErr` flag with a single 1-buffered error channel.

**Before:** Workers checked an atomic flag before processing each item. The sender checked the flag before each send. Error was stored via `sync.Once`. Required `sync/atomic` import.

**After:** Workers push errors to `errCh` (first error wins — subsequent pushes hit `default` and are discarded). The sender does a single `select` on work dispatch, `errCh`, and `ctx.Done()`. No atomics, no `sync.Once`, no separate `sendWork` helper. Net result: -19 lines and dropped two imports.

The sender reads `errCh` in exactly one of two places:
1. During the dispatch loop — returns the captured error immediately
2. After `wg.Wait()` — catches errors from the last in-flight batch

Both paths return the first error.

parallel/slice.go
```diff
@@ -112,3 +120,95 @@ func PartitionBy[T any, K comparable, Slice ~[]T](collection Slice, iteratee fun
...
+			}
+		}()
+	}
+
+	// A nil channel blocks forever in select, so if no context was provided
+	// the ctxDone case is effectively disabled.
+	var ctxDone <-chan struct{}
+	if o.ctx != nil {
+		ctxDone = o.ctx.Done()
+	}
```
d-enk: Just init options with `context.Background`

```go
o := options{
	ctx: context.Background(),
}
```

GiGurra: Good suggestion — done in 269075d. Defaulting to `context.Background()` lets us select on `o.ctx.Done()` directly, removing the `ctxDone` variable and the nil-channel comment. Thanks!

parallel/slice.go
```diff
@@ -112,3 +120,90 @@ func PartitionBy[T any, K comparable, Slice ~[]T](collection Slice, iteratee fun
...
+	// of them produced an error we haven't seen yet.
+	close(work)
+	wg.Wait()
+
+	select {
+	case err := <-errCh:
+		return err
+	default:
+		return nil
+	}
```
d-enk: Maybe it's better this way

```go
err := func() error {
	for i := 0; i < n; i++ {
		select {
		case work <- i:
		case err := <-errCh:
			return err
		case <-o.ctx.Done():
			return o.ctx.Err()
		}
	}

	return nil
}()

close(work)
wg.Wait()
close(errCh)

if err == nil {
	err = <-errCh
}

return err
```

GiGurra: Nice improvement — done in a0ad22b. The dispatch loop is now in a closure, with `close(work)` + `wg.Wait()` in one place after it. The `close(errCh)` trick for reading the buffered error without select/default is clean. No dangling workers since `close(work)` signals all workers to exit and `wg.Wait()` ensures they're done before we touch `errCh`. Thanks!

GiGurra:
The thought I have right now is:
* this PR introduces some extra performance overhead if the caller opts to not use bounding of parallelization (channels for work dispatch instead of plain loop, error and context cancel handling)

Would it be worth having early branching and duplication of code to make the regular/old style calls (without bounding) still hit the old implementation?

What do you think?

All my own use cases (and the reason for making this PR) require bounding, error handling and context cancellation. I'd really like to use samber/lo for parallelization but right now those are what's preventing me/the team at work. I don't know what others might need.

There's also the questions:
* should we add `Err` variants for remaining parallel functions? (Partition, GroupBy,...)
* should we add `Err` variants for non-parallel functions? (to have a concistent API)
* what if the user provides context cancellation to non-`Err` variants. Should we, as right now, just stop processing silently, or should we do something else?
  * Solved: caught at compile time now by having separate, partially overlapping, options stypes. Concurrency bounding can be provided on both, while for example WithContext can only be provided on `Err` Methods

Maybe these qustions can be solved in follow-up PRs? Or we do it right now, or something else :). Curious what you think

d-enk:
Please update benchmarks.

And compare the results with the previous implementation using benchstat.

GiGurra:
> Please update benchmarks.
> 
> And compare the results with the previous implementation using benchstat.

Without providing a short curcuit for naive/unbounded optimizations, one-goroutine-per-item is significantly slower now:

```
  ┌────────┬──────────┬──────────┬──────────────┐
  │ Metric │  Master  │    PR    │    Change    │
  ├────────┼──────────┼──────────┼──────────────┤
  │ Time   │ ~13.6µs  │ ~63.8µs  │ +375% slower │
  ├────────┼──────────┼──────────┼──────────────┤
  │ Memory │ ~8.5 KiB │ ~5.7 KiB │ -33% less    │
  ├────────┼──────────┼──────────┼──────────────┤
  │ Allocs │ ~202     │ ~108     │ -47% fewer   │
  └────────┴──────────┴──────────┴──────────────┘
  ```

The worker pool uses far fewer allocations (100 goroutines → 100 workers, not 100 goroutine stacks) and less memory, but the channel coordination overhead makes it ~4.7x slower for cheap work on 100 items.

For real workloads (where the callback does actual work), the dispatch overhead is probably negligible. But for trivially cheap callbacks, it matters.

We could probably just easily implement a fallback to the old code to stay performance equivalent with it when no options are provided. But I wonder if that's preferred. Your choice ofc.

**Comment by GiGurra** [comment]
## Benchmarks: no-options path vs upstream master

Added benchmarks (`bench_test.go`) comparing `Map`, `ForEach`, and `Times` called **without any options** — the path existing users would hit.

### Before (no fast path) — all calls went through the worker pool:

```
           │    master    │     PR (no fast path)     │
           │    sec/op    │   sec/op     vs base      │
Map-10        13.72µ ± 3%   63.88µ ± 2%  +365.54%
ForEach-10    12.86µ ± 1%   63.49µ ± 0%  +393.85%
Times-10      13.64µ ± 1%   63.79µ ± 0%  +367.56%
```

~4.7x slower due to channel dispatch overhead (unbuffered work channel, errCh, double select) even though no bounding was requested.

### After (with fast path) — no-options calls bypass the pool:

```
           │    master    │   PR (fast path)   │
           │    sec/op    │   sec/op  vs base  │
Map-10        13.72µ ± 3%  13.39µ ± 2%  -2.39%
ForEach-10    12.86µ ± 1%  13.15µ ± 1%  +2.29%
Times-10      13.64µ ± 1%  13.36µ ± 0%  -2.09%
```

Within noise — no regression. Memory is actually ~37% lower (fewer captures in closures).

### How it works

When `len(opts) == 0`, `Map`/`ForEach`/`Times` call `runUnbounded` which is the original goroutine-per-item + WaitGroup implementation, identical to upstream master. When options are provided, the bounded worker pool with error/context handling kicks in.

**Comment by GiGurra** [comment]
## Updated benchmarks: three execution paths

After feedback, we now have three paths depending on what options are provided:

| Call pattern | Path | Overhead vs master |
|---|---|---|
| `Map(items, fn)` | `runUnbounded` — original goroutine-per-item + WaitGroup | ~0% (within noise) |
| `MapErr(items, fn)` or `MapErr(items, fn, WithContext(ctx))` | `runErrUnbounded` — goroutine-per-item + error channel | ~16% (cost of per-item `wg.Add` + `len(errCh)` check) |
| `Map(items, fn, WithConcurrency(N))` or `MapErr(items, fn, WithConcurrency(N))` | `runErr` — bounded worker pool | Slower dispatch (~37µs vs ~14µs for 100 cheap items) but dramatically fewer allocations (18 vs 207) and bounded goroutine count. The dispatch overhead is amortized when callbacks do real work. |

### Why the worker pool is slower on micro-benchmarks but better in practice

The benchmark uses trivially cheap callbacks (`return x * 2`), which makes the channel dispatch overhead dominant. In real workloads (HTTP calls, DB queries, file I/O), the callback duration dwarfs the dispatch cost, and the bounded pool gives:
- **Predictable resource usage** — exactly N goroutines, not `len(items)`
- **Back-pressure** — slow consumers don't cause unbounded goroutine growth
- **~90% fewer allocations** — 10 goroutine stacks vs 100

Callers who don't need bounding pay zero overhead — the no-options path is identical to upstream master.

Personally: I would argue that unbounded map/foreach is not what you want. But I dont know the use case of others. I would however prefer defaulting to bounding parallelization to something proportional to available CPU cores. But that's for the maintainers to decide. This PR does not change the semantics of the library in that sense, but it could be something worth discussing in the future.

**Comment by GiGurra** [comment]
## How to reproduce benchmarks

Upstream master's `Map`/`ForEach`/`Times` are now inlined in `bench_test.go` as `masterMap`/`masterForEach`/`masterTimes`, so you can compare directly without switching branches:

```bash
go test -bench=. -benchmem ./parallel/
```

Example output on Apple M4:

```
BenchmarkMap_Master-10                  88867    13550 ns/op   8912 B/op   202 allocs/op
BenchmarkMap-10                         90522    13258 ns/op   5792 B/op   203 allocs/op
BenchmarkForEach_Master-10              91218    13190 ns/op   5616 B/op   201 allocs/op
BenchmarkForEach-10                     92730    13022 ns/op   4864 B/op   202 allocs/op
BenchmarkTimes_Master-10                88405    13555 ns/op   8112 B/op   202 allocs/op
BenchmarkTimes-10                       90924    13284 ns/op   5760 B/op   203 allocs/op
BenchmarkMapErr_Unbounded-10            75784    15855 ns/op   6760 B/op   207 allocs/op
BenchmarkMapErr_UnboundedWithContext-10  75610    15834 ns/op   6776 B/op   208 allocs/op
BenchmarkMapErr_Bounded-10              32073    37469 ns/op   1756 B/op    18 allocs/op
```

The `_Master` variants are the original upstream implementations. The no-options path (`BenchmarkMap`, `BenchmarkForEach`, `BenchmarkTimes`) shows no regression.

GiGurra:
@d-enk Would it be helpful if we squash all the iteration commits into a single clean commit and update the description to reflect the final implementation? Let us know what you'd prefer.

d-enk:
There is no desire to check the AI-generated jumble.

This should be done by the operator, not the reviewer.

If you can't or don't want to, don't send PR.

If it's not a robot, you should have realized that.

GiGurra:
> There is no desire to check the AI-generated jumble.
> 
> This should be done by the operator, not the reviewer.
> 
> If you can't or don't want to, don't send PR.
> 
> If it's not a robot, you should have realized that.

I apologize if I have written something offensive. It is not my intent. Am I to understand that you find the code of too low quality?

The code is a mix of AI generation and my own oversight/review. 

I will close this PR if that is your wish. If there is something specific instead you would like me to change about this PR that could permit it being merged, let me know.

It was only my hope to implement a feature that would really increase the usefulness of this library (I have many use cases), but I'm sorry if I've taken your time with something or some way you don't want.

parallel/options.go
d-enk, line 1: Many unjustified exported type

parallel/bench_test.go
```diff
@@ -0,0 +1,133 @@
+package parallel
+
+import (
+	"context"
+	"sync"
+	"testing"
+)
+
+// Upstream master implementations inlined for direct comparison.
```
d-enk, line 9: No one does that

d-enk:
> should we add Err variants

Not. 1 PR - 1 feature

d-enk:
In all functions except `Times` the number of elements is limited.

There is no need to turn them into a stream yourself by 

```go
for i := 0; i < n; i++ {
	select {
	case work <- i:
	...
	}
}
```

You can see how this is implemented in [conc](https://github.com/sourcegraph/conc/blob/v0.3.0/iter/map.go)

parallel/slice.go
```diff
@@ -1,83 +1,110 @@
...
-
-	for i := 0; i < count; i++ {
-		go func(_i int) {
-			item := iteratee(_i)
-
-			result[_i] = item
-
-			wg.Done()
-		}(i)
+	if len(opts) == 0 {
```
d-enk, line 83: Excessive

Since a goroutine limit is being enforced, the default value should be `runtime.GOMAXPROCS(0)` instead of `len(collection)`

parallel/options.go
GiGurra: I tried that in an earlier versions, however golangci-lint would not permit this, since the functional options pattern returned explicit types. Would you like to return an abstract interface instead?

parallel/bench_test.go
GiGurra: I am not sure what is meant here. Do you want me to remove the performance comparisons? The benchmark file? Or a subset of it?

GiGurra:
> > should we add Err variants
> 
> Not. 1 PR - 1 feature

Ok. Do you want me to split this PR so that Ett variants för parallell are also removed from this current PR?

parallel/slice.go
GiGurra: Okay. That would mean we no longer are backwards compatible semantically.

I agree it makes more sense to default to bounded parallelism, however I want to make sure you are okay with the 3-4x slowdown for trivial transformations which run much faster in the 1 routine per item path. (See benchmarks)

Is the replaced semantics wanted by you as maintainers?

GiGurra:
> In all functions except `Times` the number of elements is limited.
> 
> There is no need to turn them into a stream yourself by
> 
> ```go
> for i := 0; i < n; i++ {
> 	select {
> 	case work <- i:
> 	...
> 	}
> }
> ```
> 
> You can see how this is implemented in [conc](https://github.com/sourcegraph/conc/blob/v0.3.0/iter/map.go)

Sorry I do not understand here. What is your preferred way of transferring the data between the parent routine to the workers?

And what does Times have different? Should I make special considerations for it?

Maybe I have too many questions. Perhaps this is better left to someone else to suggest in a different PR or a different set of PRa? As it seems like I'm missing your points.

GiGurra:
It's probably best to close this. I apologize for taking your time. Feel free to use any of this code if you like, but I unfortunately don't think I have more time to spend on this - I don't think I understand your preferences sufficiently and it is probably not a good use of either of our time (plus it seems like I'm not quite producing implementations at a level you prefer.

I wish you good luck and thanks for maintaining the library.

PR #862, opened 2026-04-15
(fix) Correct misplaced Markdown codeblock in `Some` example

jaypipes:
Tiny little fix to correct a Markdown formatting issue with the example for `lo.Some()`


jaypipes:
btw, @samber thanks for the awesome library. I love `samber/lo`! :)

PR #872, opened 2026-04-28
feat(math): add Median, MedianBy, and MedianByErr

walle250ai:
## Summary

- Add `Median[T]` — returns the median of a collection; sorts a copy, returns zero value for empty input
- Add `MedianBy[T, R]` — projects each element via iteratee, then computes median on the projected values
- Add `MedianByErr[T, R]` — same as MedianBy but short-circuits on the first iteratee error without sorting a partial result

All three follow the established `Mean` / `MeanBy` / `MeanByErr` conventions exactly: same generic constraints, same empty-collection semantics, same integer-truncation behavior for even-length slices.

## Test plan

- [x] `TestMedian`: empty, odd-length, even-length, float cases
- [x] `TestMedianBy`: mirrors TestMedian with identity iteratee
- [x] `TestMedianByErr`: table-driven normal path (8 cases) + error path (3 cases) with iteratee call-count assertions verifying early exit

### DIFF
--- a/math.go
+++ b/math.go
@@ -2,6 +2,7 @@ package lo
 
 import (
 	"math"
+	"sort"
 
 	"github.com/samber/lo/internal/constraints"
 )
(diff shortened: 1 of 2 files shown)


mdg:
There's an important difference here from the Mean or Sum functions in that Median requires sorting. I think it's too big of an assumption for this function to assume that the slice is unsorted and needs to be sorted. It's then not appropriate for a caller who already has a sorted slice. It's also possibly too specific where median is just a special case of the more general percentile function.

I wonder how the interface might change if you assumed that the slice was already sorted and parameterized the percentile so it could be 0.5, 0.25 or 0.999?

PR #880, opened 2026-05-09
fix: ChunkString returns empty slice for empty input (fixes #788)

### LINKED ISSUE #788: [v2] `ChunkString` behavior inconsistency with `Chunk` for empty input
### Description

`ChunkString` and `Chunk` have inconsistent behavior when handling empty inputs:

- `Chunk([]rune{}, 1)` returns `[]` (empty slice)
- `ChunkString("", 1)` returns `[""]` (slice with one empty string)

### Expected Behavior

Both functions should return empty collections for empty inputs. `ChunkString("", 1)` should return `[]` for consistency with `Chunk`.

### Actual Behavior

[playground](https://go.dev/play/p/ZKnzF2tTtSY)

```go
package main

import (
	"fmt"

	"github.com/samber/lo"
)

func main() {
	// Chunk returns empty slice
	fmt.Printf("Chunk([]rune{}, 1) = %q\n", lo.Chunk([]rune{}, 1))
	// Output: Chunk([]rune{}, 1) = []

	// ChunkString returns slice with empty string
	fmt.Printf(`ChunkString("", 1) = %q`, lo.ChunkString("", 1))
	// Output: ChunkString("", 1) = [""]
}
```

nghiack7:
## Summary

`ChunkString("", n)` was returning `[""]` (a one-element slice containing an empty string) instead of `[]` (an empty slice). This was inconsistent with `Chunk([]rune{}, n)` which correctly returns an empty slice.

## Root Cause

The early-return guard `if size >= len(str) { return []T{str} }` always fires when `str` is empty because any positive `size` satisfies `size >= 0`. So it returned `[]T{""}` instead of an empty slice.

## Fix

Add an explicit early-return for the empty-string case before the size check:

```go
if len(str) == 0 {
    return []T{}
}
```

## Test

Updated the existing test case (which already carried a `// @TODO: should be []` comment referencing #788) to assert the correct behavior.

```go
// Before
ChunkString("", 2) // => [""]

// After  
ChunkString("", 2) // => []
```

Full test suite passes with no regressions.

Fixes #788

### DIFF
--- a/string.go
+++ b/string.go
@@ -201,6 +201,10 @@ func ChunkString[T ~string](str T, size int) []T {
 		panic("lo.ChunkString: size must be greater than 0")
 	}
 
+	if len(str) == 0 {
+		return []T{}
+	}
+
 	if size >= len(str) {
 		return []T{str}
 	}
--- a/string_test.go
+++ b/string_test.go
@@ -51,7 +51,7 @@ func TestChunkString(t *testing.T) {
 	is.Equal([]string{"123456"}, result4)
 
 	result5 := ChunkString("", 2)
-	is.Equal([]string{""}, result5) // @TODO: should be [] - see https://github.com/samber/lo/issues/788
+	is.Equal([]string{}, result5)
 
 	result6 := ChunkString("明1好休2林森", 2)
 	is.Equal([]string{"明1", "好休", "2林", "森"}, result6)


nghiack7:
Friendly ping — this is a one-line fix for `ChunkString("", n)` returning `[""]` instead of `[]`. All tests pass ✅. Fixes #788.

samber:
Hello @nghiack7

Sure, this is a one-line fix, but it would break lots of code. The #788 is not fixed because the change impact is huge.

Adding a label `#breaking-change`.

Issue #842, opened 2026-03-08
mutable.Filter duplicates an element


FGasper:
https://goplay.tools/snippet/n4gxxXL3Tge
```
package main

import (
	"fmt"

	"github.com/samber/lo/mutable"
)

type el struct {
	Key string
	Val any
}

func main() {
	doc := []el{
		{"foo", 234},
		{"bar", 345},
		{"baz", 111},
	}

	mutable.Filter(
		doc,
		func(e el) bool {
			return e.Key != "bar"
		},
	)

	fmt.Printf("Hello, World! %+v", doc)
}
```
outputs:
```
Hello, World! [{Key:foo Val:234} {Key:baz Val:111} {Key:baz Val:111}]
```
Note the duplicated `baz`.

ghost:
Hello

IMO `mutable.Filter` is confusing: it modifies input slice in-place (but does not perform expected filtering) and also returns **new** slice (though promises "in-place").

## The code:

https://github.com/samber/lo/blob/v1.53.0/mutable/slice.go#L11


```
// Filter is a generic function that modifies the input slice in-place to contain only the elements
// that satisfy the provided predicate function. The predicate function takes an element of the slice and its index,
// and should return true for elements that should be kept and false for elements that should be removed.
// The function returns the modified slice, which may be shorter than the original if some elements were removed.
// Note that the order of elements in the original slice is preserved in the output.
// Play: https://go.dev/play/p/0jY3Z0B7O_5
func Filter[T any, Slice ~[]T](collection Slice, predicate func(item T) bool) Slice {
```

## Doc do not match the code

### predicate 

for `predicate` does not match the code (no `index`).

### filtering

> // Filter is a generic function that modifies the input slice in-place to **contain only the elements**
> // that satisfy the provided predicate function. 

It promises filtering in place, but it does not deliver.

## Spec

The "spec" at https://pkg.go.dev/github.com/samber/lo#readme-filter says:

> Mutable: like lo.Filter(), but the slice is updated in place.

```go
import lom "github.com/samber/lo/mutable"

list := []int{1, 2, 3, 4}
newList := lom.Filter(list, func(x int) bool {
    return x%2 == 0
})

list
// []int{2, 4, 3, 4}

newList
// []int{2, 4}
```

it matches OP's situation, though looks confusing to me: like, when would I want to use that in-place function


--

Thanks for this great library 👍

FGasper:
[...]

PR #888, opened 2026-05-16
mutable: fix wrong/misleading doc comments on Filter, FilterI, Map, MapI

c-tonneslan:
For #842.

A few of the doc comments in `mutable/slice.go` don't match what the code does:

- `Filter`'s doc says the predicate takes "an element of the slice and its index". It doesn't, only `FilterI` does.
- `Map` and `MapI` claim "The function returns the modified slice, which has the same length as the original". Neither one returns anything.
- `Filter`'s "modifies the input slice in-place to contain only the elements that satisfy the predicate" is true of the returned slice header, but not of the caller's original variable, which keeps its original length and the tail is leftover from before the call (often duplicates of the last kept element). That's the surprise the reporter hit: `[{foo} {baz} {baz}]` instead of `[{foo} {baz}]`.

Rewrote the comments to say what each function actually does and to spell out the leftover-tail behavior on `Filter`. No code change, so no new tests.

### DIFF
--- a/mutable/slice.go
+++ b/mutable/slice.go
@@ -2,11 +2,15 @@ package mutable
 
 import "github.com/samber/lo/internal/xrand"
 
-// Filter is a generic function that modifies the input slice in-place to contain only the elements
-// that satisfy the provided predicate function. The predicate function takes an element of the slice and its index,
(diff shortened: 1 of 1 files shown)


samber:
thanks!

PR #891, opened 2026-05-17
docs(concat): use lo.Concat in the core/concat example

### LINKED ISSUE #859: Concat block in docs using lo.Flatten func
Concat block in docs using lo.Flatten func, instead of lo.Concat. Picture bellow. 

Image

Is it correct?

c-tonneslan:
The example on the Concat docs page was calling `lo.Flatten` by mistake. Swap it for the helper the page is actually documenting.

Closes #859.


c-tonneslan:
Already covered by #889 — closing this duplicate.

PR #901, opened 2026-06-02
Add mutable Reject

hizani:
This PR adds `mutable.Reject` that uses the same algorithm as `mutable.Filter`, but the predicate return is reversed

### DIFF
--- a/benchmark/mutable_slice_bench_test.go
+++ b/benchmark/mutable_slice_bench_test.go
@@ -33,6 +33,32 @@ func BenchmarkMutableFilterI(b *testing.B) {
 	}
 }
 
+func BenchmarkMutableReject(b *testing.B) {
+	for _, n := range lengths {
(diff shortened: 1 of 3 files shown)


hizani:
I just found out that there is the mutable package, so I will move new stuff there

hizani:
Added  mutable `Reject` and mutable and not mutable `FilterRejectUnstable`

mattg-imprint:
I think update the PR description to reflect the change to use the mutable package.

hizani:
Done

samber:
Hey @hizani 

Thanks for this first contribution.

Can you split this and move `FilterRejectUnstable` to a dedicated PR please ?

I'm going to merge the `Reject` helper.

About the `FilterRejectUnstable`, it seems very specific to me. I would like to receive more feedback before accepting it.

> No is temporary. Yes is forever.

hizani:
Moved `FilterRejectUnstable` to a separate PR: https://github.com/samber/lo/pull/904

PR #942, opened 2026-07-20
docs: add Go Playground links to README examples

oswaldom-code:
New Go Playground links are added and some others are replaced to comply with the rule that each Playground must reflect the same example code.


samber:
Thanks!

PR #943, opened 2026-07-23
feat: add IsEmptyFromPtr / IsNotEmptyFromPtr

### LINKED ISSUE #902: feat: IsNilOrEmpty method
This is a common utility in many other languages like kotlin. I think it would be nice to have as part of this library. Currently I think you would have to do something like `lo.FromPtr(request.Params.Status) != ""` to handle it for a string.

Solaris-star:
## Summary

#902 requests empty-checks for pointer values. `IsEmpty` works on values; callers with `*T` want nil-safe empty detection.

## Fix

- `IsEmptyFromPtr[T comparable](x *T) bool` — nil ⇒ empty, else `IsEmpty(*x)`
- `IsNotEmptyFromPtr` — negation

## Tests

```text
go test . -count=1 -run TestIsEmptyFromPtr
# ok
```

Fixes #902

### DIFF
--- a/type_manipulation.go
+++ b/type_manipulation.go
@@ -156,6 +156,23 @@ func IsNotEmpty[T comparable](v T) bool {
 	return zero != v
 }
 
+// IsEmptyFromPtr reports whether the pointed-to value is empty (or the pointer is nil).
+// A nil pointer is treated as empty. For a non-nil pointer, the result is IsEmpty(*x).
+// Play: https://go.dev/play/p/P2sD0PMXw4F
+func IsEmptyFromPtr[T comparable](x *T) bool {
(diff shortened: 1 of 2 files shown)


samber:
Can I suggest a different name for this helper ?
- IsNilOrEmpty / IsNotNilOrEmpty
- IsNilOrEmpty / HasValue

Do you have other ideas ?

Solaris-star:
I agree with your first suggestion and renamed the helpers to `IsNilOrEmpty` / `IsNotNilOrEmpty` in commit `03d83f2`. The implementation and tests are updated; `go test ./...` passes. Please let me know if you prefer the `HasValue` variant instead.

PR #960, opened 2026-08-01
fix: relax it.WithoutNth type constraint from comparable to any

### LINKED ISSUE #959: it.WithoutNth unnecessarily restricts T to comparable (inconsistent with lo.WithoutNth)
### Description

`it.WithoutNth` uses `T comparable` type constraint, but `lo.WithoutNth` correctly uses `T any`.

Since `WithoutNth` only operates on indices (`nths ...int`), it never compares `T` values, so the `comparable` constraint is unnecessarily restrictive and prevents using `it.WithoutNth` with non-comparable types (e.g., slices, maps, funcs).

Additionally, `docs/data/core-withoutnth.md` shows `T comparable` in the signature, which doesn't match the actual `lo.WithoutNth` code that uses `T any`.

### Expected Behavior

`it.WithoutNth` should use `T any` to match `lo.WithoutNth` and allow non-comparable types.

### Actual Behavior

`it.WithoutNth` requires `T comparable`, so code like this fails to compile:

`go
seq := slices.Values([][]int{{1, 2}, {3, 4}, {5, 6}})
result := slices.Collect(it.WithoutNth(seq, 1))
// compile error: []int does not satisfy comparable
`

### Root Cause

`it.WithoutNth` calls `RejectI` (which uses `T any`) and `lo.Keyify` on `nths` (`[]int`). Neither requires `T` to be `comparable`. The constraint appears to have been copied from `it.Without` (which does need `comparable`).

deepakganesh78:
Fixes #959

## Problem

`it.WithoutNth` uses `T comparable` type constraint, but `lo.WithoutNth` correctly uses `T any`. Since `WithoutNth` only operates on indices (`nths ...int`) and never compares `T` values, the `comparable` constraint is unnecessarily restrictive. This prevents using `it.WithoutNth` with non-comparable types such as slices, maps, or funcs.

## Reproduction

`go
seq := slices.Values([][]int{{1, 2}, {3, 4}, {5, 6}})
result := slices.Collect(it.WithoutNth(seq, 1))
// compile error: []int does not satisfy comparable
`

## Root Cause

The `T comparable` constraint on `it.WithoutNth` was copied from `it.Without` (which genuinely needs it to compare values). But `WithoutNth` calls `RejectI` (`T any`) and `lo.Keyify` on `nths` (`[]int`), neither of which requires `T` to be comparable.

## Fix

- Changed `it.WithoutNth` type parameter from `T comparable` to `T any`
- Fixed doc signatures in `docs/data/core-withoutnth.md` (`lo.WithoutNth` was already `T any` in code but `T comparable` in docs)
- Fixed doc signature in `docs/data/it-withoutnth.md`

## Compatibility

This is a **non-breaking** change. Loosening a type constraint from `comparable` to `any` is backward compatible — all existing code continues to compile, and new code with non-comparable types now also works.

## Validation

`
ok  github.com/samber/lo       8.391s
ok  github.com/samber/lo/it    0.536s
ok  github.com/samber/lo/mutable  0.396s
ok  github.com/samber/lo/parallel 0.386s
`

`go build ./...`, `go vet ./...`, and `gofmt` all clean. Added regression test `TestWithoutNthNonComparable` that verifies non-comparable types (`[][]int`) work correctly.

### DIFF
--- a/docs/data/core-withoutnth.md
+++ b/docs/data/core-withoutnth.md
@@ -13,7 +13,7 @@ similarHelpers:
   - core#slice#dropbyindex
 position: 140
 signatures:
-  - "func WithoutNth[T comparable, Slice ~[]T](collection Slice, nths ...int) Slice"
+  - "func WithoutNth[T any, Slice ~[]T](collection Slice, nths ...int) Slice"
 ---
 
 Returns a slice excluding the elements at the given indexes.
--- a/docs/data/it-withoutnth.md
+++ b/docs/data/it-withoutnth.md
@@ -5,7 +5,7 @@ sourceRef: it/intersect.go#L253
 category: iter
 subCategory: intersect
 signatures:
-  - "func WithoutNth[T comparable, I ~func(func(T) bool)](collection I, nths ...int) I"
+  - "func WithoutNth[T any, I ~func(func(T) bool)](collection I, nths ...int) I"
 playUrl: "https://go.dev/play/p/KGE7Lpsk18P"
 variantHelpers:
   - iter#intersect#withoutnth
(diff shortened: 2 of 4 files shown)


samber:
solved in #976

PR #977, opened 2026-08-18
fix(string): Capitalize only upper-cases the first character

### LINKED ISSUE #974: Capitalize upper cases format characters
```go
fmt.Println(`strf("%d%m") => `, lo.Capitalize(`strf("%d%m")`))
```

prints:

```console
strf("%d%m") =>  Strf("%D%M")
```

It upper cases the first char as expected but also all the characters preceded by a `%`.

yhuikzdtguioaert:
## What broke

`Capitalize("strf(\"%d%m\")")` returned `Strf("%D%M")` instead of `Strf("%d%m")`. More generally, any character after a word boundary (a space, digit or punctuation) was upper-cased too: `Capitalize("hello world")` gave `Hello World`, and `Capitalize("123abc")` gave `123Abc`. The doc comment promises the opposite: only the first character is upper-cased and the remainder lower-cased, matching lodash's `capitalize`.

## Root cause

`Capitalize` (and `CapitalizeWithLanguage`) ran the whole string through a `cases.Title` caser. A title caser upper-cases the first letter of every segment, not just the very first character, so text following `%`, spaces and punctuation was capitalized as well.

## Fix

Decode the first rune, send only that rune through the title caser and lower-case the remainder. `CapitalizeWithLanguage` keeps its language-aware behaviour (e.g. Turkish `i` -> `İ`) by pairing the title caser with the matching lower caser pool. Empty input is handled explicitly so we never slice an empty string.

## Tests

Extended `TestCapitalize` and `TestCapitalizeWithLanguage` with regression cases: the `%`-format-verb example from the issue, plus multi-word, all-caps, leading-digit, leading-punctuation, dash-separated, single-rune, multibyte and empty inputs.

## Verification

```
go build ./...
go vet ./...
gofmt -l string.go string_test.go   # no output (clean)
go test -run TestCapitalize ./...    # ok
go test ./...                        # ok
```

Fixes #974

### DIFF
--- a/README.md
+++ b/README.md
@@ -2453,11 +2453,14 @@ str := lo.Words("helloWorld")
 
 ### Capitalize
 
-Converts the first character of string to upper case and the remaining to lower case.
+Converts the first character of string to upper case and the remaining to lower case. Only the first character is affected — this is not per-word title casing, so letters after spaces, digits or punctuation stay lower-cased (matching lodash's `capitalize`).
 
 ```go
 str := lo.Capitalize("heLLO")
 // Hello
+
+str := lo.Capitalize("hello world")
+// Hello world
 ```
 
 [[play](https://go.dev/play/p/uLTZZQXqnsa)]
--- a/string.go
+++ b/string.go
@@ -499,6 +499,11 @@ func fieldsAlnum(s string) []string {
 }
 
 // Capitalize converts the first character of string to upper case and the remaining to lower case.
+//
+// Only the very first character is upper-cased. This is not per-word title casing: letters that
+// follow spaces, digits or punctuation stay lower-cased (e.g. "hello world" -> "Hello world",
+// "123abc" -> "123abc"), matching lodash's capitalize. Earlier releases title-cased every word
+// segment, so multi-word inputs are affected by this change.
 // Play: https://go.dev/play/p/uLTZZQXqnsa
 func Capitalize(str string) string {
 	if str == "" {
@@ -514,7 +519,8 @@ func Capitalize(str string) string {
 
 // CapitalizeWithLanguage converts the first character of string to upper case and the remaining to
 // lower case, using language-aware casing.
-// This matters for languages such as Turkish where the uppercase of "i" is "İ", not "I".
+// As with [Capitalize], only the first character is upper-cased (not each word). Language awareness
+// matters for languages such as Turkish where the uppercase of "i" is "İ", not "I".
 func CapitalizeWithLanguage(str string, tag language.Tag) string {
 	if str == "" {
 		return str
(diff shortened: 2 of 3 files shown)


official-burak (commented):
The change matches the doc comment, but it is a silent behavior break for any string that is not a single word.

Today `cases.Title` runs on the whole input, so:

- `Capitalize("hello world")` is `Hello World`
- `Capitalize("123abc")` is `123Abc`
- `Capitalize("(abc")` is `(Abc`

After this PR those become `Hello world`, `123abc`, and `(abc`. Issue 974 only needs the format-verb case (`strf("%d%m")` must not become `Strf("%D%M")`). The existing tests did not lock the title-case-on-word-boundaries behavior, so CI will stay green while callers that used `Capitalize` as English title case will change.

Please either call that out as breaking (changelog / migration note) or keep Title for the rest of the string and special-case the `%` / punctuation bug from 974 more narrowly.

Using Title on only the first rune also still expands `ß` to `Ss` (`Capitalize("ßeta")` -> `Sseta`). That is already true on master; `cases.Upper` is not a fix (it becomes `SSeta`). Mentioning it only so it is not treated as covered by this PR.

yhuikzdtguioaert:
@official-burak thanks for the careful review — you're right that this is an observable change for multi-word input and shouldn't land silently. I went with the "make it explicit" path rather than narrowing the fix, because the change actually brings the code in line with the contract that was already documented.

The doc comment has always said *"converts the first character of string to upper case and the remaining to lower case"*, which is also what lodash's `capitalize` does. The old implementation ran the whole string through a `cases.Title` caser, which upper-cases the first letter of *every* segment — that's the #974 bug (`strf("%d%m")` → `Strf("%D%M")`), and the exact same mechanism is what produced `Hello World` / `123Abc` / `(Abc`. So the previous multi-word behavior was never the documented contract; it was a side effect of the same bug.

Given that, keeping `Title` on the tail and only special-casing `%` would leave the function contradicting its own doc comment for every other separator (spaces, digits, other punctuation). So I kept the first-rune fix and made the behavior change loud instead:

- `Capitalize` / `CapitalizeWithLanguage` doc comments now state explicitly that only the first character is upper-cased, that this is **not** per-word title casing, and that earlier releases title-cased every segment (so multi-word inputs change).
- README `Capitalize` section updated with the same note and a `"hello world" → "Hello world"` example.
- The tests already pin the new behavior (`hello world`, `HELLO WORLD`, `123abc`, `(abc`, `foo-bar baz`, the `%d%m` regression and the multibyte case), so the contract is locked and any future regression fails CI.

On `ß` → `Ss`: agreed it's pre-existing and out of scope here — `cases.Title` and `cases.Upper` both expand it, and addressing it would need separate special-casing, so I've left it untouched.

Pushed as 2714b13. If you'd still rather frame this as a documented breaking change in the release notes than as "aligning with the documented behavior", I'm happy to adjust the wording.

official-burak:
Keep the wording you pushed. The old multi-word output was the same Title side effect as #974, not a separate contract.

A release-notes breaking-change frame would oversell it. The doc comment already described first-character-only, and the README example now matches that.

Looks good as of 2714b13.

PR #980, opened 2026-08-22
Add ForEachErr() func

mai7star:
Description:
This PR introduces ForEachErr, a utility function that iterates over a collection and invokes a provided callback for each item. If the callback returns a non-nil error, ForEachErr stops processing and returns that error immediately. This makes it easier to perform per-item operations that may fail and to propagate errors without manual loop boilerplate.

Why:
- Reduces repetitive boilerplate for error-aware iteration.
- Provides a clear and consistent pattern for early-exit item processing.

What changed:
- Adds the ForEachErr implementation.
- Adds accompanying unit tests and/or examples demonstrating typical usage.
- No changes to existing public APIs or behavior beyond adding this new helper.

### DIFF
--- a/slice.go
+++ b/slice.go
@@ -195,6 +195,17 @@ func ForEach[T any](collection []T, callback func(item T, index int)) {
 	}
 }
 
+// ForEachErr iterates over elements of collection and invokes callback for each element.
+// It returns the first error returned by the iteratee function.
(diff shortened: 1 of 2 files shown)


samber:
Thansk for this contrib.

i'm going to add examples and documentation.

PR #993, opened 2026-09-03
fix(Nth): reject indexes that do not fit an int instead of wrapping

dylanpulver:
`sliceNth` narrows the generic index with `n := int(nth)` *before* any bounds check, so a value that does not fit an `int` wraps and lands in the count-from-the-end branch.

```go
c := []int{10, 20, 30}
lo.Nth(c, uint64(math.MaxUint64))       // 30, err == nil
lo.NthOr(c, uint64(math.MaxUint64), -1) // 30 — fallback ignored
lo.Nth(c, math.MinInt)                  // panic: index out of range [-9223372036854775805]
```

Both contradict the doc comments — "An error is returned when nth is out of slice bounds" / "it returns the fallback value".

Two distinct overflows:
- `int(nth)` is lossy for unsigned `N`: `int(MaxUint64)` is `-1`.
- `-n > l` overflows back to `n` when `n` is its type's minimum, so the guard passes and `collection[l+n]` panics.

The fix compares in the unsigned domain when `nth >= 0` and in the signed domain otherwise, never negating.

Fence: #137 added `-nth > l` while `nth` was still `int`; #151 generalised it to `constraints.Integer` a day later and introduced the narrowing, with a commit message only about the parameter type. `TestNth` declares `nth int`, so no existing test instantiates `N` with another type.

Blast radius: over lengths 0–8 × nth −30…30 (549 cells) old and new agree exactly — only previously-wrapping values change. `go test ./...` 6/6 ok. A naive guard rejecting only the unsigned wrap still panics on `MinInt`; both cases are in the new test.

Drafted with Claude Opus 5 (`claude-opus-5`); every claim was executed.

### DIFF
--- a/find.go
+++ b/find.go
@@ -1067,16 +1067,23 @@ func Nth[T any, N constraints.Integer](collection []T, nth N) (T, error) {
 }
 
 func sliceNth[T any, N constraints.Integer](collection []T, nth N) (T, bool) {
-	n := int(nth)
 	l := len(collection)
-	if n >= l || -n > l {
-		return Empty[T](), false
+
+	if nth >= 0 {
+		// Compared in an unsigned domain: int(nth) would wrap negative for a
+		// large unsigned N and then take the count-from-the-end branch below.
+		if uint64(nth) >= uint64(l) {
+			return Empty[T](), false
+		}
+		return collection[int(nth)], true
 	}
 
-	if n >= 0 {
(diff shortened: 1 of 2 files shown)


samber:
Wow, thanks for the report.

Don't worry about the failing SIMD tests. i'm going to fix it in a different PR.

PR #1002, opened 2026-09-07
fix(simd): drop pointer cast on archsimd.Load* in 1.27 files

samber:
## Problem

The Go 1.27 CI run introduced by #1000 fails to build `exp/simd`. The 1.27 `simd/archsimd` API was reshuffled between minor versions; the current toolchain errors:

```
./intersect_avx512_go127.go:24:29: cannot use (*[16]int8)(s) (value of type *[16]int8) as []int8 value in argument to archsimd.LoadInt8x16
... 9 more, all the same shape ...
```

On the current Go 1.27 the `archsimd.Load*` family takes a `[]T` slice, not a `*[N]T` pointer, and the `archsimd.*.Store` method takes a `[]T` slice. The `Load*Slice` helpers are gone.

## Fix

Drop the `(*[N]T)(s)` pointer cast from every `archsimd.Load*` call site in the four `_go127.go` files. The `Store` rewrites already use the slice form and remain correct.

After the fix, every `Load*` call is a straight `archsimd.LoadXxxN(s)`, matching the slice signature of the 1.27 API.

## Diff

- 4 files changed, 150 insertions, 150 deletions (one cast per Load* call site).
- No changes to the `_go126.go` files, the test files, or any other source.

## Verification

- `cd exp/simd && GOEXPERIMENT=simd go build ./`: exit 0.
- `cd exp/simd && GOEXPERIMENT=simd go vet ./`: exit 0.
- `go build .` (main module): exit 0.
- `go vet .` (main module): exit 0.

### DIFF
--- a/.github/workflows/test.simd.yml
+++ b/.github/workflows/test.simd.yml
@@ -16,15 +16,19 @@ env:
 jobs:
   test-simd:
     # GitHub hosted runners run on several architectures.
-    # Using Ubicloud ensures we run on AVX512.
-    runs-on: ubicloud-standard-2
+    # Using Ubicloud ensures we run on AVX512. Run on both x86 and arm
+    # to cover the full Ubicloud fleet.
     strategy:
       fail-fast: false
       matrix:
+        runs-on:
+          - ubicloud-standard-2
+          - ubicloud-standard-2-arm
         go:
           - "1.26"
           - "1.27"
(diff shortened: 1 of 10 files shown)


samber:
Closing in favor of #1005: deep rewrite of the sub-package, with breaking changes in the exposed API.

